PR #5, opened 2018-02-07
Fix spelling mistake for 'Cannon Lake'

bsurmanski:
See:
https://www.intel.com/content/www/us/en/design/products-and-solutions/processors-and-chipsets/platform-codenames.html
https://en.wikipedia.org/wiki/Cannon_Lake_(microarchitecture)


gchatelet:
Thank you for your contribution!

PR #8, opened 2018-02-07
fix typo in readme

cwalv:


gchatelet:
Thx for your contribution but the README is correct.
Either you know the code has been compiled with AVX (in which case you don't need to read it at runtime) **or** you read it at runtime.

I recognize that the documentation may not be clear enough, I'll update it. Thank you!

cwalv:
Hmm .. what would happen if the code was compiled to use the AVX extensions
but the CPU doesn't support it?  Are you basically saying "this was defined
at compile time, so we're going to assume that it exists"?

On Thu, Feb 8, 2018 at 7:33 AM, Guillaume Chatelet <notifications@github.com

PR #25, opened 2018-02-13
detect intel sgx and smx cpu features for the x86 arch

xqp:
with this patch the lib detect [sgx](https://en.wikipedia.org/wiki/Software_Guard_Extensions) and smx

### DIFF
--- a/src/cpuinfo_x86.c
+++ b/src/cpuinfo_x86.c
@@ -102,6 +102,9 @@ static void ParseCpuId(const uint32_t max_cpuid_leaf, X86Info* info) {
   features->bmi2 = IsBitSet(leaf_7.ebx, 8);
   features->vpclmulqdq = IsBitSet(leaf_7.ecx, 10);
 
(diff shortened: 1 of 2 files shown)


src/cpuinfo_x86.c
```diff
@@ -102,6 +102,9 @@ static void ParseCpuId(const uint32_t max_cpuid_leaf, X86Info* info) {
   features->bmi2 = IsBitSet(leaf_7.ebx, 8);
   features->vpclmulqdq = IsBitSet(leaf_7.ecx, 10);
 
+  features->smx = IsBitSet(leaf_1.ecx, 6);
+  features->sgx = IsBitSet(leaf_7.ebx, 2);
```
gchatelet: Do you mind moving these a few lines above?
`sgx` before `bmi1` and `smx` before `aes`

gchatelet:
Thank you for this PR :+1:

PR #27, opened 2018-02-16
Adding cross compilation

### LINKED ISSUE #13: Add continuous integration for ARM / aarch64 / mipsel

gchatelet:
Fixes #13.


scripts/run_integration.sh
```diff
@@ -0,0 +1,173 @@
+#!/bin/bash
+
+readonly SCRIPT_FOLDER=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
+readonly PROJECT_FOLDER="${SCRIPT_FOLDER}/.."
+readonly ARCHIVE_FOLDER=~/archives
```
Mizux: Can't you use /var/cache/cpu_features or /tmp (and change it in .travis.yml also) ?

scripts/run_integration.sh
```diff
@@ -0,0 +1,173 @@
...
+}
+
+function unpackifnotexists() {
+  mkdir -p ${ARCHIVE_FOLDER}
+  cd ${ARCHIVE_FOLDER}
+  local URL=$1
+  local DESTINATION=`pwd`/$2
+  if [[  ! -d "${DESTINATION}" ]] ; then
+    local ARCHIVE_NAME=`echo ${URL} | sed 's/.*\///'`
+    test -f ${ARCHIVE_NAME} || wget ${URL}
```
Mizux: I would use -q, otherwise in travis.ci log you could have tons of progress bar lines (and reach the 4Mb limit).
cf https://api.travis-ci.org/v3/job/341333288/log.txt

.travis.yml
```diff
@@ -1,16 +1,81 @@
...
+      TOOLCHAIN=CODESCAPE
+      TARGET=mips-mti-linux-gnu
+      QEMU_ARCH=DISABLED
 
 script:
   - cmake --version
-  - cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON -H. -Bcmake_build
-  - cmake --build cmake_build --target all
-  - CTEST_OUTPUT_ON_FAILURE=1 cmake --build cmake_build --target test
+  - bash -e -x ./scripts/run_integration.sh
```
Mizux: if you look at your log 75% seems to be cross toolchain/qemu build
-> could you create an "install" script and use the "install:" key instead ?
Since travis only provides "folding" for each build steps....

Otherwise need to scroll at bottom to see the cmake (useful) traces...

scripts/run_integration.sh
```diff
@@ -0,0 +1,173 @@
...
+      installqemuifneeded
+      QEMU="${QEMU_INSTALL}/bin/qemu-${QEMU_ARCH} ${QEMU_ARGS}"
+    fi
+  else
+    QEMU=""
+  fi
+  # Run tests
+  for test_binary in ${BUILD_DIR}/test/*_test; do ${QEMU} ${test_binary}; done
+  # Run demo program
+  ${QEMU} ${BUILD_DIR}/list_cpu_features
```
Mizux: I would run the demo first to get the qemu config, then run the test knowing the qemu "environment"...

.travis.yml
gchatelet: Good idea. I'll do it later though.

scripts/run_integration.sh
gchatelet: I'm not sure I understand what you mean. They both use QEMU (which may be empty).

Issue #34, opened 2018-04-26
Overwriting windows syscalls


Leandros:
Windows, the Win32 API in specific, has a function called `ReadFile`, `OpenFile` and `CloseFile`. You define all three in filesystem.c, non-namespaced, and therefore overriding the Win32 syscalls for everybody who links with your library. This is a huge oversight and cost me a couple of hours of trying to figure out why my programs all start to crash when reading files out of a sudden.

https://github.com/google/cpu_features/blob/7eca4a7a399e8c96a6dacf09c0c59d558c62f45a/src/filesystem.c#L24-L30

gchatelet:
Hi @Leandros 

Thx a lot for reporting this: I wasn't aware obviously.
Let's move the conversation on PR https://github.com/google/cpu_features/pull/35

PR #35, opened 2018-04-26
Explicitly namespace every extern identifier

Leandros:
To avoid identifier collisions and fix https://github.com/google/cpu_features/issues/34, I moved every extern identifier into their own "namespace", by prefixing them with `CpuFeatures_` (or `CpuFeatures_ClassName_`, for example for the `StringView`: `CpuFeatures_StringView_`).

I'm aware this is a huge change, but it's, unfortunately, required since C has no explicit namespaces. The other option would be to migrate to C++ and use actual namespaces.
I only did a search & replace for the identifiers, the change is likely to violate some 80 column limits and would need to be reformatted at some points. 

Tested and compiles on macOS. 

**Edit:** I also just realized I haven't updated the tests.


Leandros:
Signed.

gchatelet:
Thx a lot for the PR @Leandros .
Do you mind updating the tests and running `git clang-format` ?

gchatelet:
Also, can we fix only the filesystem API for now and not touch capabilities and stringview?

I believe that `struct StringView` has more chance to collide with existing code than the functions working on it.

Leandros:
> I believe that struct StringView has more chance to collide with existing code than the functions working on it.

That doesn't matter, since that collision would only occur if you'd actually include the `string_view.h` header into your own code. The StringView functions, on the other hand, are in the global namespace, and will be accessible by any program or library which is also linked into the same application as this library. And I'd say, a function like `Push` or `Back` can very likely appear in some C or C++ programs.

tbodt:
If the StringView functions are only called by the library, and wouldn't be called by any users of the library, they could be marked as `__attribute__((visibility("hidden")))` or the equivalent on windows (not sure what that is).

Leandros:
That's a good idea, which would reduce the amount of change required.

gchatelet:
> That doesn't matter, since that collision would only occur if you'd actually include the string_view.h header into your own code. The StringView functions, on the other hand, are in the global namespace, and will be accessible by any program or library which is also linked into the same application as this library. And I'd say, a function like Push or Back can very likely appear in some C or C++ programs.

@Leandros you're right. I forgot `C` does not mangle function names and that's indeed [ODR violation](https://en.wikipedia.org/wiki/One_Definition_Rule).

I think the visibility way is promising (and adds less clutter to the code). IIRC MSVC does not export symbols if not marked `__declspec(dllexport)` so it should already work properly with MSVC. Can you confirm you used GCC or Clang?

Meanwhile, I'll work on a fix based on @tbodt suggestion (thx @tbodt!)

Leandros:
I used MSVC. `__declspec(dllexport)` only works for DLLs, not for static libraries.

Leandros:
Tests should be fixed now. You'd have to run `clang-format`, though. Seems my clang-format 5.0.1 is to old to recognize the config file.

gchatelet:
As for clang-format, it may be because of a missing line feed at the end of the file.
I submitted a fix. Can you patch the PR and try again?

Leandros:
Lovely, that seems to work. I've fixed the failing tests and re-formated all files.

gchatelet:
Fantastic! Thank you for your contribution, your time and for bearing with me :) My `C` is quite rusted as you can see :)

PR #36, opened 2018-05-01
Add Power support

RashmicaG:
Thoughts, opinions, questions and queries are welcome.


src/cpuinfo_ppc.c
```diff
@@ -0,0 +1,359 @@
...
+    {{0, PPC_FEATURE2_ARCH_3_00}, "arch_3_00", &set_arch300},
+    {{0, PPC_FEATURE2_HAS_IEEE128}, "ieee128", &set_ieee128},
+    {{0, PPC_FEATURE2_DARN}, "darn", &set_darn},
+    {{0, PPC_FEATURE2_SCV}, "scv", &set_scv},
+    {{0, PPC_FEATURE2_HTM_NO_SUSPEND}, "htm-no-suspend", &set_htm_no_suspend},
+};
+static const size_t kConfigsSize = sizeof(kConfigs) / sizeof(CapabilityConfig);
+
+static bool HandlePPCLine(const LineResult result,
+                              PPCPlatformStrings* const strings) {
```
gchatelet: Do you mind running `git clang-format`?

RashmicaG: How do you have clang-format set up? When I try to run it I get 
```
Error reading /home/rashmica/working/cpu_features/.clang-format: Invalid argument
YAML:4:4: error: unknown key '...'
...
   ^

```
I've pushed the changes that this config gave me - figured it should be close to what you're looking for?
``` clang-format -style=google -dump-config > .clang-format```

File: src/cpuinfo_ppc.c [hunk 2]
```diff
@@ -0,0 +1,359 @@
...
+static char model[MAX_LINE_SIZE];
+static char machine[MAX_LINE_SIZE];
+static char cpu[MAX_LINE_SIZE];
+static const PPCPlatformStrings kEmptyPPCPlatformStrings = {
+	.platform = platform, .model = model, .machine = machine, .cpu = cpu};
+
+PPCPlatformStrings GetPPCPlatformStrings(void) {
+  PPCPlatformStrings strings = kEmptyPPCPlatformStrings;
+
+  FillProcCpuInfoData(&strings);
```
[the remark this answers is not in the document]
RashmicaG: This is a bit gross sorry. 

At first I did want to avoid dynamic memory allocation, but then I realised that most of the information needed is can be deduced from the feature flags*. Hence not including the `PPCPlatformStrings` struct in the  `PPCInfo` struct to keep `GetPPCInfo()` memcpy/malloc free by avoiding strings.

I was going to only parse `/proc/cpuinfo` once in `GetPPCPlatformStrings` and have a static 'PPCPlatformStrings' struct that would be copied everytime  `GetPPCPlatformStrings` was called. But then I figured I should talk to you about that, and didn't tidy up properly :woman_facepalming:  

Do you think it makes sense to only parse `/proc/cpuinfo` once in `GetPPCPlatformStrings`? Or should I leave that behaviour as is because a process shouldn't repeatedly call it?  

*The only information that I think you could gain from the `strings` struct that you can't from the flags would be the actual hardware architecture if your process was running in compat mode (ie running with POWER7 instructions on a POWER8 architecture).

src/cpuinfo_ppc.c
gchatelet: It may be because of a missing line feed at the end of `.clang-format`. I submitted a fix, can you patch the PR and try again?

src/cpuinfo_ppc.c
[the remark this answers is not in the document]
gchatelet: I actually discussed this internally when designing the library. We decided that `GetXXXInfo()` should do all the work always from scratch and not try to reuse data. This removes the problem of thread synchronization entirely. On the performance side this is not a big deal and the client can always cache the result if need be.

There's no need for dynamic allocation, just return a struct containing an array of char, it will be copied for you when the struct is returned.

include/cpuinfo_ppc.h
```diff
@@ -0,0 +1,142 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef CPU_FEATURES_INCLUDE_CPUINFO_PPC_H_
+#define CPU_FEATURES_INCLUDE_CPUINFO_PPC_H_
+
+#include "cpu_features_macros.h"
+#include "internal/hwcaps.h"
+
+#define MAX_LINE_SIZE 100
```
gchatelet: This will leak out of the header and possibly conflict (or replace) other definitions.
Just set the number in the struct directly.

Also 100 is probably too much. What would be a good size here? 32? 64?

include/cpuinfo_ppc.h
```diff
@@ -0,0 +1,142 @@
...
+
+typedef struct {
+  PPCFeatures features;
+} PPCInfo;
+
+// This function is guaranteed to be malloc, memset and memcpy free.
+PPCInfo GetPPCInfo(void);
+
+typedef struct {
+  char platform[MAX_LINE_SIZE];
```
gchatelet: Please add a comment after each string that it is 0 terminated.

include/internal/hwcaps.h
```diff
@@ -61,13 +61,73 @@ CPU_FEATURES_START_CPP_NAMESPACE
...
-  uint32_t hwcaps;
-  uint32_t hwcaps2;
+  unsigned long hwcaps;
+  unsigned long hwcaps2;
 } HardwareCapabilities;
 
 HardwareCapabilities GetHardwareCapabilities(void);
 
+typedef struct {
+  char platform[100];
```
gchatelet: Same here and below. Add a comment about 0 terminated strings.
Possibly reduce size to 32 or 64 (what makes more sense).

src/hwcaps.c
```diff
@@ -152,6 +158,17 @@ HardwareCapabilities GetHardwareCapabilities(void) {
...
 }
 
+PlatformType kEmptyPlatformType;
+
+PlatformType GetPlatformType(void) {
+  PlatformType type = kEmptyPlatformType;
+  char *platform = (char *)GetHardwareCapabilitiesFor(AT_PLATFORM);
+  char *base_platform = (char *)GetHardwareCapabilitiesFor(AT_BASE_PLATFORM);
+
+  if (platform != NULL) strcpy(type.platform, platform);
```
gchatelet: You need to check that `platform` fits into `type.platform`, if not copy up to `sizeof(type.platform) - 1` and zero terminate the string.

I'll add function to do this in a separate PR.

test/hwcaps_for_testing.cc
```diff
@@ -29,4 +32,10 @@ HardwareCapabilities GetHardwareCapabilities(void) {
   return *g_hardware_capabilities;
 }
 
+void SetPlatformTypes(const char* platform, const char* base_platform) {
+  strcpy(g_platform_types->platform, platform);
```
gchatelet: Ditto.

include/cpuinfo_ppc.h
RashmicaG: Good point, I'll do that and use 64.

src/cpuinfo_ppc.c
RashmicaG: Yup, no issues running git clang-format now.

src/cpuinfo_ppc.c
```diff
@@ -0,0 +1,356 @@
...
+static const size_t kConfigsSize = sizeof(kConfigs) / sizeof(CapabilityConfig);
+
+static bool HandlePPCLine(const LineResult result,
+                          PPCPlatformStrings* const strings) {
+  StringView line = result.line;
+  StringView key, value;
+  if (CpuFeatures_StringView_GetAttributeKeyValue(line, &key, &value)) {
+    if (CpuFeatures_StringView_HasWord(key, "platform")) {
+      CpuFeatures_StringView_CopyString(value, strings->platform,
+                                        value.size + 1);
```
gchatelet: `CpuFeatures_StringView_CopyString` third argument is the size of the destination buffer so it should be
```
CpuFeatures_StringView_CopyString(value, strings->platform, sizeof(strings->platform));
```

Please fix the following lines as well.

src/hwcaps.c
gchatelet: Actually you can use the following snippet to get the job done:
```
if (platform != NULL)
  CpuFeatures_StringView_CopyString(str(platform), type.platform, sizeof(type.platform));
```

test/hwcaps_for_testing.cc
gchatelet: Same here and below, use `CpuFeatures_StringView_CopyString`

src/cpuinfo_ppc.c
RashmicaG: woops, not sure how I got that confused!

gchatelet:
Looks good to me. Thank you so much for your contribution @RashmicaG :)

PR #37, opened 2018-05-22
without cpu_features_macros.h in the install tree, the library cannot

psteinb:
be used after `make install`


gchatelet:
Thx! and sorry for the lag...

PR #42, opened 2018-08-30
Small changes made according to static analyzer results

fuzun:
I did not want to to change type of feature variables since it seems that they are getting accessed directly by other projects but could not find any other solution. Making them bool would probably break portability. Other option might be defining TRUE as -1 then modifying related parts.

Feature variable returning functions (Get...FeaturesEnumValue) are kept intact as automatic casting should not cause any problem. And because variables are all aligned, there should be no warning because of this casting in compilers other than VS. (I did not test though).


fuzun:
I signed it!

gchatelet:
So unfortunately this PR is changing the API and may break client code.
I'm willing to pull everything in but the `int` -> `unsigned int`modification (we would also need to modify the other platforms arm, aarch64, power ...)

fuzun:
I have realized that bit_utils.h is internal so made some changes there instead of structs. Still not the best but I could not think any other solution without changing structures.

include/internal/bit_utils.h
```diff
@@ -22,8 +22,11 @@
 
 CPU_FEATURES_START_CPP_NAMESPACE
 
-inline static bool IsBitSet(uint32_t reg, uint32_t bit) {
-  return (reg >> bit) & 0x1;
+#define TRUE -1
```
gchatelet: Let's not go this route. Redefining TRUE is opening a can of worms.
Let's just drop this commit for now.

gchatelet:
Thx a lot for your patience and for contributing to `cpu_features`.
Can you open a bug about the bool correctness issue so I keep it on my radar?

fuzun:
Sure!

PR #48, opened 2019-01-15
Declare CMAKE_SYSTEM_NAME/SYSTEM for Linaro builds

gchatelet:
As well as fixing typo and shebang.


scripts/test_integration.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
Mizux: #!/usr/bin/env bash

gchatelet: Done

PR #49, opened 2019-01-15
Per arch build and inlining of cpuid_x86.

gchatelet:
This PR is about:
 - Modifying `CMakeLists.txt` to allow per arch build instead of building everything on all architectures,
 - Update tests so they use a preprocessor definition to short-circuit implementation
 - Inline files `src/cpuid_x86_clang_gcc.c` and `src/cpuid_x86_msvc.c` into `src/cpuinfo_x86.c`


CMakeLists.txt
```diff
@@ -21,38 +21,48 @@ option(BUILD_SHARED_LIBS "Build library as shared." OFF)
...
+    list(APPEND _SRCS src/cpuinfo_aarch64.c)
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+    list(APPEND _HDRS include/cpuinfo_x86.h)
+    list(APPEND _HDRS include/internal/cpuid_x86.h)
+    list(APPEND _SRCS src/cpuinfo_x86.c)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(powerpc|ppc)")
+    list(APPEND _HDRS include/cpuinfo_ppc.h)
+    list(APPEND _SRCS src/cpuinfo_ppc.c)
+else()
+  message(SEND_ERROR "Unsupported architectures ${CMAKE_SYSTEM_PROCESSOR}")
```
Mizux: FATAL_ERROR ? i.e. is there any interest in continuing the processing of the CMakeLists.txt

CMakeLists.txt
```diff
@@ -21,38 +21,48 @@ option(BUILD_SHARED_LIBS "Build library as shared." OFF)
...
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^mips")
+    list(APPEND _HDRS include/cpuinfo_mips.h)
+    list(APPEND _SRCS src/cpuinfo_mips.c)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^arm")
+    list(APPEND _HDRS include/cpuinfo_arm.h)
+    list(APPEND _SRCS src/cpuinfo_arm.c)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^aarch64")
+    list(APPEND _HDRS include/cpuinfo_aarch64.h)
+    list(APPEND _SRCS src/cpuinfo_aarch64.c)
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
```
Mizux: x86 ? if library only support x86_64 why file are not suffixed by x86_64 ?

src/hwcaps.c
```diff
@@ -101,12 +91,19 @@ static uint32_t GetElfHwcapFromGetauxval(uint32_t hwcap_type) {
   dlclose(libc_handle);
   return ret;
 }
-#endif  // defined(HWCAPS_ANDROID_MIPS_OR_ARM)
+#elif defined(CPU_FEATURES_OS_LINUX_OR_ANDROID)
+// On Regular Linux we simply use getauxval.
+#include <dlfcn.h>
+#include <sys/auxv.h>
```
Mizux: May you should check via cmake that "sys/auxv.h" exists.
see https://cmake.org/cmake/help/latest/module/CheckIncludeFile.html

test/CMakeLists.txt
```diff
@@ -54,9 +56,12 @@ target_link_libraries(linux_features_aggregator_test all_libraries)
 add_test(NAME linux_features_aggregator_test COMMAND linux_features_aggregator_test)
 ##------------------------------------------------------------------------------
 ## cpuinfo_x86_test
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
 add_executable(cpuinfo_x86_test cpuinfo_x86_test.cc ../src/cpuinfo_x86.c)
+target_compile_definitions(stack_line_reader PUBLIC CPU_FEATURES_MOCK_CPUID_X86)
```
Mizux: good target ? if yes you should separate it from `cpuinfo_x86_test` stuff

gchatelet: Good catch

CMakeLists.txt
gchatelet: I edited the condition to include all x86 flavors at once.

CMakeLists.txt
```diff
@@ -32,44 +32,55 @@ set(_HDRS
...
-  include/internal/string_view.h
-  src/linux_features_aggregator.c
-  src/filesystem.c
-  src/hwcaps.c
-  src/stack_line_reader.c
-  src/string_view.c
 )
 
+macro(add_linux_detection)
+  if(NOT (APPLE OR UNIX))
```
Mizux: if(NOT UNIX)
In CMake APPLE is a Unix-like system i.e. APPLE is a subset of UNIX
see: https://cmake.org/cmake/help/latest/variable/UNIX.html

gchatelet: Even better

PR #51, opened 2019-01-16
Update macros to detect mips64 and differentiate between x86 32/64.

gchatelet:


include/cpu_features_macros.h
```diff
@@ -37,7 +47,11 @@
 #define CPU_FEATURES_ARCH_ANY_ARM
 #endif
 
-#if defined(__mips__)
+#if defined(__mips64)
+#define CPU_FEATURES_ARCH_MIPS64
+#endif
+
+#if defined(__mips__) && !defined(__mips64)  // mips64 also declares __mips__
 #define CPU_FEATURES_ARCH_MIPS
```
Mizux: should you define CPU_FEATURES_ARCH_MIPS_32 here instead
then
#if defined(CPU_FEATURES_ARCH_MIPS_32) || defined(CPU_FEATURES_ARCH_MIPS_64)
#define CPU_FEATURES_ARCH_MIPS
#endif

to be consistent with X86 flavors...

Issue #47, opened 2019-01-12
Add an easy way to migrate from the Android NDK


gchatelet:
Provide a drop in replacement for the [NDK's cpu-features.h](https://android.googlesource.com/platform/ndk/+/master/sources/android/cpufeatures/cpu-features.h)

PR #53, opened 2019-01-17
Export GetArmCpuId function

gchatelet:
This is in preparation of adding [ndk_compat](https://github.com/google/cpu_features/issues/47)

### DIFF
--- a/include/cpuinfo_arm.h
+++ b/include/cpuinfo_arm.h
@@ -15,6 +15,7 @@
 #ifndef CPU_FEATURES_INCLUDE_CPUINFO_ARM_H_
 #define CPU_FEATURES_INCLUDE_CPUINFO_ARM_H_
 
+#include <stdint.h>  // uint32_t
 #include "cpu_features_macros.h"
 
 CPU_FEATURES_START_CPP_NAMESPACE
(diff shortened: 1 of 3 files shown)


include/cpuinfo_arm.h
```diff
@@ -15,6 +15,7 @@
 #ifndef CPU_FEATURES_INCLUDE_CPUINFO_ARM_H_
 #define CPU_FEATURES_INCLUDE_CPUINFO_ARM_H_
 
+#include <stdint.h>  // uint32_t
```
Mizux: this header is C99, is it ok for you ?
see: https://en.cppreference.com/w/c/types/integer

gchatelet: Yes.
This is stated in the description of the project that we are expecting C99.

Mizux: FYI if we target cmake 3.1 (or above)
then we could add:
```cmake
set_target_properties(tgt PROPERTIES
  C_STANDARD 99
  C_STANDARD_REQUIRED ON)
```
src: https://cmake.org/cmake/help/v3.1/prop_tgt/C_STANDARD.html
https://cmake.org/cmake/help/v3.1/command/set_target_properties.html

PR #57, opened 2019-01-18
Add mips release number

gchatelet:


Mizux:
You also have R2 and R5, should this PR contains the three ?
https://github.com/google/cpu_features/blob/26133d3b620c2c27f31d571efd27371100f891e9/include/internal/hwcaps.h#L57-L59

gchatelet:
@Mizux as discussed there were a bug in the implementation and R2 and R5 where not existent as a Linux hardware capability. I'll just add the bare minimum to get the NDK work then.

gchatelet:
This PR is part of solving #47 .

PR #58, opened 2019-01-18
Add x86 missing feature detections for ndk_compat

gchatelet:


Mizux:
You also have R2 and R5, should this PR contains the three ?
https://github.com/google/cpu_features/blob/26133d3b620c2c27f31d571efd27371100f891e9/include/internal/hwcaps.h#L57-L59

include/cpuinfo_x86.h
```diff
@@ -140,6 +144,10 @@ typedef enum {
   X86_SMX,
   X86_SGX,
   X86_CX16,
+  X86_SHA,
+  X86_POPCNT,
+  X86_MOVBE,
+  X86_RDRND,
```
Mizux: prefer to use `RDRAND` like the ANDROIR macro.
note: wikipedia seems to use `RdRand` https://fr.wikipedia.org/wiki/RdRand

ndk_compat/cpu-features.c
```diff
@@ -128,11 +128,11 @@ static void android_cpuInit(void) {
   if (info.features.sse4_2) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_SSE4_2;
   if (info.features.avx) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_AVX;
   if (info.features.avx2) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_AVX2;
-    // if (info.features.) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_MOVBE;
-    // if (info.features.) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_AES_NI;
-    // if (info.features.) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_RDRAND;
-    // if (info.features.) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_POPCNT;
-    // if (info.features.) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_SHA_NI;
+  if (info.features.movbe) g_cpuFeatures |= ANDROID_CPU_X86_FEATURE_MOVBE;
```
Mizux: nit: I would have organised this check in the same order than the `X86FeaturesEnum` list.
i.e. `aes` at top then here: `sha`, `popcnt`, `movbe`, `rdrand`

src/cpuinfo_x86.c
```diff
@@ -145,12 +145,16 @@ static void ParseCpuId(const uint32_t max_cpuid_leaf, X86Info* info) {
 
   features->smx = IsBitSet(leaf_1.ecx, 6);
   features->cx16 = IsBitSet(leaf_1.ecx, 13);
+  features->movbe = IsBitSet(leaf_1.ecx, 22);
+  features->popcnt = IsBitSet(leaf_1.ecx, 23);
+  features->rdrnd = IsBitSet(leaf_1.ecx, 30);
```
Mizux: rdrand should be after f16c to write them in order

include/cpuinfo_x86.h
gchatelet: I've used the naming from https://en.wikipedia.org/wiki/CPUID for the field in the `X86Features` struct and I wanted to make sure the enums and the field names are spelled the same way - since they need to stay in sync. What do you think?

ndk_compat/cpu-features.c
gchatelet: I reordered to use the same order as in the NDK.

src/cpuinfo_x86.c
gchatelet: Done

include/cpuinfo_x86.h
gchatelet: So if you don't mind I'll proceed and submit this patch as-is.
Let me know what you think by commenting here and I'll fix it in a subsequent PR.

PR #69, opened 2019-03-19
Convert XGETBV to equivalent byte form in GetXCR0Eax

natanbc:
[osxcross](https://github.com/tpoechtrager/osxcross) gives the following error, which also happens with [regular gcc on OS X](https://github.com/asmjit/asmjit/issues/78):
```
cpu_features/src/cpuinfo_x86.c:44:no such instruction: `XGETBV'
```

Changing to the byte form fixes that error, while still working properly.


natanbc:
I signed it!

gchatelet:
thx a lot @natanbc !

PR #74, opened 2019-06-15
Avoid if-clause in CpuFeatures_StringView_IsEquals()

psiegl:


psiegl:
Less if/else-clauses mean less testing effort required. Well yes, it is debatable if moving if-clause into longer boolean expression is better. From my perspective: yes ;-)

gchatelet:
> Less if/else-clauses mean less testing effort required.
How come?

psiegl:
> How come?
I get your point: the code does just a conversion from an if/else-clause to a boolean expression and is therefore equal. Feel free to merge or close it.

gchatelet:
Let's keep it as is then. Thx!

PR #77, opened 2019-06-20
Add x86 CpuIdEx function to pass inputs in ecx register

fexolm:
(required E.g for leaf4)

P.S. I can try to add support for leaf4 soon.

### DIFF
--- a/src/cpuinfo_x86.c
+++ b/src/cpuinfo_x86.c
@@ -33,9 +33,9 @@
 
 #include <cpuid.h>
 
-Leaf CpuId(uint32_t leaf_id) {
+Leaf CpuIdEx(uint32_t leaf_id, int ecx) {
   Leaf leaf;
-  __cpuid_count(leaf_id, 0, leaf.eax, leaf.ebx, leaf.ecx, leaf.edx);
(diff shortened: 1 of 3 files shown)


src/cpuinfo_x86.c
```diff
@@ -70,16 +70,24 @@ uint32_t GetXCR0Eax(void) { return _xgetbv(0); }
 #error "Unsupported compiler, x86 cpuid requires either GCC, Clang or MSVC."
 #endif
 
+static Leaf CpuId(uint32_t leaf_id) {
```
gchatelet: Is this used?

fexolm: Yes, but it can be replaced with CpuIdEx(leafid, 0). Then I can remove old CpuId function and rename CpuIdEx into CpuId.

fexolm: I wasn't sure what solution fits best here, so decided to make new function.

gchatelet: Then I think it shouldn't be `static`

fexolm: But it used only in `cpuinfo_x86.c`.

gchatelet: Ha my bad, sorry about the confusion.

PR #81, opened 2019-06-23
Fix ARMv6 reporting on Linux vs Android

psiegl:


src/cpuinfo_arm.c
```diff
@@ -102,7 +102,11 @@ static bool HandleArmLine(const LineResult result, ArmInfo* const info,
       const StringView digits =
           CpuFeatures_StringView_KeepFront(value, IndexOfNonDigit(value));
       info->architecture = CpuFeatures_StringView_ParsePositiveNumber(digits);
-    } else if (CpuFeatures_StringView_IsEquals(key, str("Processor"))) {
+    } else if (CpuFeatures_StringView_IsEquals(key, str("Processor"))
+               || CpuFeatures_StringView_IsEquals(key, str("model name")) ) {
+      // Android reports this in a non-Linux standard "Processor" but sometimes
+      // also in "model name", Linux reports it only in "model name"
+      // see RaspberryPiZero (Linux) vs InvalidArmv7 (Android) test-cases
```
digit-google: As far as I know, this only depends on the version of the kernel being used. Newer Linux kernels all report this in model_name for ARM, but some older ones used to report it in Processor (also older ARM kernels were mostly custom forks maintained by OEMs with all kinds of crazy patches like that). And Android devices used to ship with a lot of dated kernels :)

gchatelet: So is this `model name` or `model_name`?. Or both? :-D

psiegl: as far as I am aware `model name`

digit-google: Sorry my bad, it is "model name" of course :)

PR #83, opened 2019-06-29
Use enum properly to directly access kConfig struct

psiegl:


src/cpuinfo_aarch64.c
```diff
@@ -113,29 +116,15 @@ int GetAarch64FeaturesEnumValue(const Aarch64Features* features,
     case AARCH64_CRC32:
       return features->crc32;
     case AARCH64_LAST_:
+    default:
```
gchatelet: Remove the default case here and in all the switches to allow the compiler to warn about missing enum case.

psiegl: fixed

src/cpuinfo_aarch64.c
```diff
@@ -119,23 +122,8 @@ int GetAarch64FeaturesEnumValue(const Aarch64Features* features,
...
-      return "sha2";
-    case AARCH64_CRC32:
-      return "crc32";
-    case AARCH64_LAST_:
-      break;
-  }
-  return "unknown feature";
+  if(value >= kConfigsSize)
+    return "unknown feature";
+  else
```
gchatelet: Ha also can you remove the `else` here and below?
Sorry I should have mentioned it in the previous review...

psiegl: fixed

gchatelet:
Thx!

Issue #4, opened 2018-02-07
SSSE3, SSE4.1, SSE4.2 not detected on non-AVX CPUs


Myriachan:
ParseCpuId is coded to require the XMM XCR0 bit in order to recognize SSSE3, SSE4.1, and SSE4.2, but this is not the right way to detect it.  Machines without AVX will not have XCR0 at all, and will fail this check.

In the absence of XCR0, on x86-32, the proper way to detect SSE is to ask the operating system.  (On x86-64, just assume yes.)  In Windows, that's IsProcessorFeaturePresent(PF_XMMI_INSTRUCTIONS_AVAILABLE).  In Linux, unfortunately, it's parsing /proc/cpuinfo.  In macOS and iOS (simulator), it's sysctlbyname on hw.optional.sse, though you could also just assume yes.

Something could be said that the number of x86 machines without SSE in the CPU and OS is about zero now--it depends on how far back you want your library to support.

aprelev:
Ping.

WilliamTambellini:
I confirm the bug:

processor       : 1
vendor_id       : GenuineIntel
cpu family      : 6
model           : 44
model name      : Intel(R) Xeon(R) CPU           X5650  @ 2.67GHz
stepping        : 2
cpu MHz         : 2660.000
cache size      : 256 KB
physical id     : 0
siblings        : 4
core id         : 1
cpu cores       : 4
apicid          : 1
initial apicid  : 1
fpu             : yes
fpu_exception   : yes
cpuid level     : 11
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht pni ssse3 cx16 sse4_1 sse4_2 popcnt aes hypervisor lahf_lm ida arat dtherm
clflush size    : 64
cache_alignment : 64
address sizes   : 40 bits physical, 48 bits virtual

and list_cpu_features does nt report much: 

$ list_cpu_features
arch            : x86
brand           : Intel(R) Xeon(R) CPU           X5650  @ 2.67GHz
family          :   6 (0x06)
model           :  44 (0x2C)
stepping        :   2 (0x02)
uarch           : INTEL_WSM
flags           : aes,cx16,popcnt

WilliamTambellini:
[...]

PR #92, opened 2019-07-23
Support CherryTrail that doesn't indicate xmm os support

psiegl:
There was already a debate in https://github.com/google/cpu_features/issues/4 if cpus that do not set the XCR0 properly related to the xmm os support should be supported or not.
I stumpled across a CherryTail Atom that explicitly has this issue: XCR0 bit for xmm not set, but providing SSE* register set.

Now, on the one hand the debate above depicted "won't fix" as there won't be that many cpus with this issue. This is odd, because on the front page of https://github.com/google/cpu_features it is clearly depicted that a dev needs to take care about poor hardware implementations (example with SNB and AVX) in his/her own code. But in this case cpu_features tries to take control in case the bit in XCR0 for xmm os support is (not) set. I'd suggest to make the XCR0 xmm os support bit explicit to the dev so he/she can decide as well?
Hence, this patch removes so far the burden of not properly indicating SSE* on cpus that didn't set the XCR0 register properly, but still support SSE*.

### DIFF
--- a/src/cpuinfo_x86.c
+++ b/src/cpuinfo_x86.c
@@ -96,12 +96,6 @@ static bool HasMask(uint32_t value, uint32_t mask) {
   return (value & mask) == mask;
 }
 
-// Checks that operating system saves and restores xmm registers during context
-// switches.
-static bool HasXmmOsXSave(uint32_t xcr0_eax) {
-  return HasMask(xcr0_eax, MASK_XMM);
-}
-
 // Checks that operating system saves and restores ymm registers during context
 // switches.
 static bool HasYmmOsXSave(uint32_t xcr0_eax) {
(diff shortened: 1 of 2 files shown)


Mizux:
I feel it overkill to purely suppress the XCR0 xmm check, why don't creating a white list where we disable the XCR0 check if we now these architectures are known to no set this bit ?

something like this ?
```c
bool force_sse_os_support = false;
// ATOM platform support sse but don't set the XCR0 bit...
if (GetX86Microarchitecture(&info) == X86Microarchitecture::INTEL_ATOM_SMT) {
  force_sse_os_support = true;
}
....
const bool have_sse_os_support = HasXmmOsXSave(xcr0_eax) || force_sse_os_support;
```

psiegl:
Sounds good to me. I just can not test it right now, but code looks good. Would you mind creating a patch?

gchatelet:
This is now fixed by #135

PR #94, opened 2019-11-12
[NFC] Use a tree structure in list_cpu_features

gchatelet:
This is in preparation to include cache hierarchy in the dumped data.

### DIFF
--- a/src/utils/list_cpu_features.c
+++ b/src/utils/list_cpu_features.c
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This program dumps current host data to the standard output.
+// Output can be text or json if the `--json` flag is passed.
+
+#include <assert.h>
+#include <stdarg.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+
+// The type of the nodes in the tree.
+typedef enum {
+  TNT_INVALID,
+  TNT_INT,
+  TNT_MAP,
+  TNT_MAP_ENTRY,
+  TNT_ARRAY,
+  TNT_ARRAY_ELEMENT,
+  TNT_STRING,
```
digit-google: nit: Shouldn't this use the TVT_ prefix instead?

src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+  if (ptr) BA = (BumpAllocator){.ptr = ptr, .size = size};
+  return BA;
+}
+
+// Update the available memory left in the BumpAllocator.
+static void* BA_Bump(BumpAllocator* BA, size_t size) {
+  assert(BA->size >= size);
+  void* ptr = BA->ptr;
+  BA->size -= size;
+  BA->ptr += size;
```
digit-google: This allocation doesn't consider alignment requirements, so this is going to return un-aligned addresses that will be cast to pointers to aligned structs (e.g. TreeValueType*), which will crash on certain platforms. A quick fix for this is to ensure the final allocated size is a multiple of 8.

src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+  TreeValueType type;
+  unsigned integer;
+  const char* string;
+  struct TreeValue* value;
+  struct TreeValue* next;
+} TreeValue;
+
+// Allocates a node inside a BumpAllocator.
+static TreeValue* BA_TreeValue(BumpAllocator* BA, TreeValueType type) {
+  TreeValue* TV = (TreeValue*)BA_Bump(BA, sizeof(TreeValue));
```
digit-google: nit: using capital letters for variables is unconventional. Can you use |ba| and |tv| instead?

src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+  assert(TV);
+  TV->type = type;
+  return TV;
```
digit-google: This leaves most fields in uninitialized state, which is a recipe for subtle bugs. I'd recommend clearing all fields before returning to catch them as soon as possible in the future.

src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+
+// Allocates a formatted string inside a BumpAllocator.
+static TreeValue* CreatePrintfString(BumpAllocator* BA, const char* format,
+                                     ...) {
+  va_list arglist;
+  va_start(arglist, format);
+  char* const ptr = BA->ptr;
+  const int written = vsnprintf(ptr, BA->size, format, arglist);
+  va_end(arglist);
+  if (written < 0 || written >= BA->size) return NULL;
```
digit-google: I would recommend aborting the program immediately if this fails. There is little the program can do in this case, and it is likely that the NULL will never be checked by 100% of callers.

src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+// We use a bump allocator to allocate strings and nodes of the tree,
+// Memory is not intented to be reclaimed.
+typedef struct {
+  char* ptr;
+  size_t size;
+} BumpAllocator;
+
+// Allocate a buffer of size `size`.
+static BumpAllocator BA_Create(size_t size) {
+  char* const ptr = (char*)malloc(size);
```
digit-google: For something that simple (i.e. a 64 KiB buffer used by a single program), I would recommend just using a global static buffer and BumpAllocator variable. This would ensure everything is clear on startup (see comments below), and would simply all functions to not need to |BA| parameter.

src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+} TreeValueType;
+
+// The node in the tree.
+typedef struct TreeValue {
+  TreeValueType type;
+  unsigned integer;
+  const char* string;
+  struct TreeValue* value;
+  struct TreeValue* next;
+} TreeValue;
```
digit-google: nit: It may make more sense to call this a Dictionary than a Tree, since it doesn't have the regular left/right children nodes and performance characteristics of trees.

gchatelet:
Thx a lot for the comments @digit-google I really appreciate it.
I'll address them as a separate PR, I'll make sure to add you as reviewer next time :)

src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+  va_end(arglist);
+  if (written < 0 || written >= BA->size) return NULL;
+  return CreateConstantString(BA, (char*)BA_Bump(BA, written));
+}
+
+static TreeValue* CreateString(BumpAllocator* BA, const char* value) {
+  return CreatePrintfString(BA, "%s", value);
+}
+
+// Allocates a map entry node inside a BumpAllocator.
```
digit-google: nit: technically, this add an entry to |map|. The BumpAllocator is an implementation detail, and it would probably makes more sense to put it as the last parameter.

src/utils/list_cpu_features.c
```diff
@@ -30,7 +36,176 @@
...
+                 Aarch64Features, AARCH64_LAST_)
+#elif defined(CPU_FEATURES_ARCH_MIPS)
+DEFINE_ADD_FLAGS(GetMipsFeaturesEnumValue, GetMipsFeaturesEnumName,
+                 MipsFeatures, MIPS_LAST_)
+#elif defined(CPU_FEATURES_ARCH_PPC)
+DEFINE_ADD_FLAGS(GetPPCFeaturesEnumValue, GetPPCFeaturesEnumName, PPCFeatures,
+                 PPC_LAST_)
+#endif
+
+// Prints a json string with characters escaping.
```
digit-google: nit: where is the character escaping? If not ready yet, please add a TODO

PR #95, opened 2019-11-12
Address comments in https://github.com/google/cpu_features/pull/94

gchatelet:
Addressing comments in https://github.com/google/cpu_features/pull/94.
@digit-google can you take a look?


gchatelet:
> Nice :) Hurray for C struct literals btw, they rock ;)

Thx for the review and thx for the great comments. The code is much better now :)

PR #98, opened 2019-12-14
Use a getter function to avoid manual work for future to-be-added cpu…

psiegl:
To avoid modifying the Get*FeaturesEnumValue() function, one can use a setter function.


psiegl:
@gchatelet any chance of getting this code in or any comments on it?

psiegl:
@gchatelet added 3 test-cases to the ARM architecture. One of them tests for out of bound.

gchatelet:
Thx !

PR #99, opened 2020-01-05
Differentiate between different Lake uarch

### LINKED ISSUE #97: cpu microarchitecture is not correct
CPU-Z shows that CPU's code name is `Coffe Lake`, but this code's output is `Kaby Lake`.

Moxeja:
Issue #97 is caused by the fact that the Family and Model numbers of Kaby Lake and Coffee Lake CPUs are the same. The only way I could see to work around that is by using their Stepping values to differentiate them. 

According to https://en.wikichip.org/wiki/intel/microarchitectures/kaby_lake and https://en.wikichip.org/wiki/intel/microarchitectures/coffee_lake, these are:

Kaby Lake Stepping values: 9,
Coffee Lake Stepping values: 10, 11, 12, 13

I'm quite new to Git and C so please give me any necessary feedback to improve this commit.
Fixes #97 

Edit:
I've also included code changes to properly detect Cannon Lake, Ice Lake and Whiskey Lake processors.

### DIFF
--- a/src/cpuinfo_x86.c
+++ b/src/cpuinfo_x86.c
@@ -689,10 +689,27 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
       case CPUID(0x06, 0x5E):
         // https://en.wikipedia.org/wiki/Skylake_(microarchitecture)
         return INTEL_SKL;
+      case CPUID(0x06, 0x66):
+        // https://en.wikipedia.org/wiki/Cannon_Lake_(microarchitecture)
+        return INTEL_CNL;
+      case CPUID(0x06, 0x7E):
+        // https://en.wikipedia.org/wiki/Ice_Lake_(microprocessor)
+        return INTEL_ICL;
       case CPUID(0x06, 0x8E):
+        switch (info->stepping) {
+          case 9:  return INTEL_KBL;  // https://en.wikipedia.org/wiki/Kaby_Lake
+          case 10: return INTEL_CFL;  // https://en.wikipedia.org/wiki/Coffee_Lake
+          case 11: return INTEL_WHL;  // https://en.wikipedia.org/wiki/Whiskey_Lake_(microarchitecture)
+          default: return X86_UNKNOWN;
+        }
       case CPUID(0x06, 0x9E):
-        // https://en.wikipedia.org/wiki/Kaby_Lake
-        return INTEL_KBL;
+        if (info->stepping > 9) {
+          // https://en.wikipedia.org/wiki/Coffee_Lake
+          return INTEL_CFL;
+        } else {
+          // https://en.wikipedia.org/wiki/Kaby_Lake
+          return INTEL_KBL;
+        }
       default:
         return X86_UNKNOWN;
     }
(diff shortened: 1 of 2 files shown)


src/cpuinfo_x86.c
```diff
@@ -689,10 +689,25 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
...
+        // https://en.wikipedia.org/wiki/Cannon_Lake_(microarchitecture)
+        return INTEL_CNL;
+      case CPUID(0x06, 0x7E):
+        // https://en.wikipedia.org/wiki/Ice_Lake_(microprocessor)
+        return INTEL_ICL;
       case CPUID(0x06, 0x8E):
+        if (info->stepping > 10) {
+          // https://en.wikipedia.org/wiki/Whiskey_Lake_(microarchitecture)
+          return INTEL_WHL;
+        }
```
Mizux: Please add to 0x06, 0x8E
```cpp
if (9 == info->stepping) {
           // https://en.wikipedia.org/wiki/Kaby_Lake
           return INTEL_KBL;
} else if (10 == info->stepping) {
           // https://en.wikipedia.org/wiki/Coffee_Lake
           return INTEL_CFL;
} else if (11 == info->stepping) {
          // https://en.wikipedia.org/wiki/Whiskey_Lake_(microarchitecture)
          return INTEL_WHL;  
} else {
  return X86_UNKNOWN;
}
```

ref:
* https://en.wikichip.org/wiki/intel/microarchitectures/kaby_lake#CPUID
Family 6 Model 142 Stepping 9

* https://en.wikichip.org/wiki/intel/microarchitectures/coffee_lake#CPUID
Family 6 Model 142 Stepping 10

* https://en.wikichip.org/wiki/intel/microarchitectures/whiskey_lake#CPUID
Family 6 Model 142 Stepping 11

Moxeja: @Mizux Is this ok?

src/cpuinfo_x86.c
```diff
@@ -689,10 +689,33 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
...
         // https://en.wikipedia.org/wiki/Skylake_(microarchitecture)
         return INTEL_SKL;
+      case CPUID(0x06, 0x66):
+        // https://en.wikipedia.org/wiki/Cannon_Lake_(microarchitecture)
+        return INTEL_CNL;
+      case CPUID(0x06, 0x7E):
+        // https://en.wikipedia.org/wiki/Ice_Lake_(microprocessor)
+        return INTEL_ICL;
       case CPUID(0x06, 0x8E):
+        if (info->stepping == 9) {
```
gchatelet: Can you use a switch statement instead?
```
switch(info->stepping) {
  case 9: return INTEL_KBL; // https://en.wikipedia.org/wiki/Kaby_Lake
  case 10: return INTEL_CFL; // https://en.wikipedia.org/wiki/Coffee_Lake
  case 11: return INTEL_WHL; // https://en.wikipedia.org/wiki/Whiskey_Lake_(microarchitecture)
  default: return X86_UNKNOWN;
}
```

Moxeja:
@gchatelet  I've made the requested changes.

PR #100, opened 2020-01-06
Add INTEL_WHL and INTEL_ICL to uarch get name function

Moxeja:
Follow up from #99. Forgot to add INTEL_WHL and INTEL_ICL to the GetX86MicroarchitectureName function.


Mizux:
good catch ! many thanks again for the effort !

PR #103, opened 2020-01-28
Fixes #102 - wrong macros

### LINKED ISSUE #102: Compile-time checks not OK?
Hi, when following [your readme](https://github.com/google/cpu_features#checking-compile-time-flags) my compiler complained and I can't disagree with it: Your code
```CPU_FEATURES_COMPILED_X86_AVX || features.avx;```
after CPP expands to
```defined(__AVX__) || features.avx;```
and that doesn't compile (or is there a trick to get the CPP-command `defined` to become executable?).

--> Wouldn't it be better to do something like this in `cpu_features_macros.h`:
```c
#ifdef (__AVX__)
#define CPU_FEATURES_COMPILED_X86_AVX 1
#else 
#define CPU_FEATURES_COMPILED_X86_AVX 0
#endif 
```
instead of the current
```c
#define CPU_FEATURES_COMPILED_X86_AVX defined(__AVX__)
```
to make your sample code work?

Or should one use your sample code in a specific / different way than by simply gcc-ing it?

gchatelet:


Mizux:
@googlebot I consent.

PR #104, opened 2020-01-29
[NFC] Use Designated Initializers

gchatelet:
This is a mechanical change. This is to prevent error in argument ordering.
@fexolm Can you provide me with the source of the data you used for traceability?

### DIFF
--- a/src/cpuinfo_x86.c
+++ b/src/cpuinfo_x86.c
@@ -131,344 +131,859 @@ static int IsVendor(const Leaf leaf, const char* const name) {
 
 static const CacheLevelInfo kEmptyCacheLevelInfo;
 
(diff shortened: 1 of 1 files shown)


fexolm:
@gchatelet I've used [Intel SDM](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf) page 313 (if you mean that).

fexolm:
Also you may take a look at https://github.com/intel-go/cpuid/blob/master/cpuid_amd64.go#L467

gchatelet:
This is what I meant yes, thank you !

Issue #106, opened 2020-02-05
Fatal error: 'cpu-features.h' file not found


leshchenko:
I downloaded lib and copied it to my project directory. My CMakeLists.txt :

```cmake
 cmake_minimum_required(VERSION 3.6)

 add_subdirectory(cpu_features)

 add_library(
     myLib
     SHARED
     app/src/main/cpp/test.cpp
)
 
 link_libraries(myLib cpu_features)
```

My test.cpp file:

```cpp
#include "test.h"
#include <cpu-features.h>

```
I have `10: fatal error: 'cpu-features.h' file not found` error.

Please tell me what I'm doing wrong. 
Thanks.

Mizux:
you must use `target_link_libraries()` ... or move up the `link_libraries()` IMHO

leshchenko:
@Mizux Thanks for the quick reply. Could you please help me in case I want to use `cpu-features.h` from https://github.com/google/cpu_features/tree/master/ndk_compat ?

Mizux:
in this cas e you should have something like this I guess:
```cmake
...
add_subdirectory(cpu_features/ndk_compat)

 add_library(
     myLib
     SHARED
     app/src/main/cpp/test.cpp
)
  target_link_libraries(myLib ndk_compat)

leshchenkoEk:
@Mizux Unfortunately it doesn't work

gchatelet:
@leshchenkoEk can you confirm that it works now?

Mizux:
Should be fixed by ma patch see the comment in the PR...

PR #107, opened 2020-02-05
CMake: Add missing alias for cpu_feature

Mizux:
For customers using add_subdirectory() approach

related to #106


gchatelet:
Thx @Mizux !

PR #111, opened 2020-03-11
Support x86 FMA4 and SSE4A features

gadoofou87:


Mizux:
can be found in https://en.wikipedia.org/wiki/CPUID#EAX=80000001h:_Extended_Processor_Info_and_Feature_Bits

src/cpuinfo_x86.c
```diff
@@ -1097,9 +1097,11 @@ static void ParseCpuId(const uint32_t max_cpuid_leaf, X86Info* info) {
     features->ssse3 = IsBitSet(leaf_1.ecx, 9);
     features->sse4_1 = IsBitSet(leaf_1.ecx, 19);
     features->sse4_2 = IsBitSet(leaf_1.ecx, 20);
+    features->sse4a = IsBitSet(leaf_7.ecx, 6);
```
Mizux: IIRC, leaf_7 is EAX=7
while here we want EAX=80000001h ie leaf_80000001 ?

gadoofou87: I really apologize for that. I made a mistake due to inattention.

src/cpuinfo_x86.c
```diff
@@ -1043,6 +1043,8 @@ static void ParseLeaf4(const int max_cpuid_leaf, CacheInfo* info) {
 static void ParseCpuId(const uint32_t max_cpuid_leaf, X86Info* info) {
   const Leaf leaf_1 = SafeCpuId(max_cpuid_leaf, 1);
   const Leaf leaf_7 = SafeCpuId(max_cpuid_leaf, 7);
+  const Leaf leaf_80000000 = CpuId(2147483648);
+  const Leaf leaf_80000001 = SafeCpuId(leaf_80000000.eax, 2147483649);
```
gchatelet: Can we stick to hex values here?
```
const Leaf leaf_80000000 = CpuId(0x80000000);
const Leaf leaf_80000001 = SafeCpuId(leaf_80000000.eax, 0x80000001);
```

src/cpuinfo_x86.c
```diff
@@ -1043,6 +1043,8 @@ static void ParseLeaf4(const int max_cpuid_leaf, CacheInfo* info) {
 static void ParseCpuId(const uint32_t max_cpuid_leaf, X86Info* info) {
   const Leaf leaf_1 = SafeCpuId(max_cpuid_leaf, 1);
   const Leaf leaf_7 = SafeCpuId(max_cpuid_leaf, 7);
+  const Leaf leaf_80000000 = CpuId(2147483648);
```
gchatelet: This code should be conditionally executed since it's only valid for AMD CPUs.
To be honest I'd rather have a separate `ParseExtraAMDCpuId` dealing with AMD only features.

src/cpuinfo_x86.c
Mizux: done ;)

src/cpuinfo_x86.c
Mizux: maybe here for the call https://github.com/google/cpu_features/blob/eb168a2da2e14a85834b98c3961eee909dc618f9/src/cpuinfo_x86.c#L1135-L1137

Mizux: @gadoofou87  on my way to patch this PR...

Mizux:
@googlebot I consent.

src/cpuinfo_x86.c
```diff
@@ -1135,6 +1158,9 @@ X86Info GetX86Info(void) {
   if (IsVendor(leaf_0, "GenuineIntel") || IsVendor(leaf_0, "AuthenticAMD")) {
```
gchatelet: Here let's instantiate a struct like this:
```
struct OsSupport {
  bool have_sse_os_support;
  bool have_avx_os_support;
};
```

Then pass the struct by pointer to `ParseCpuId` so it can fill it, and pass the struct by value to `ParseExtraAMDCpuId` so it can read it. This way we don't redo the work (each call to cpuid is ~100 cycles)

Mizux: done

src/cpuinfo_x86.c
```diff
@@ -1039,17 +1039,23 @@ static void ParseLeaf4(const int max_cpuid_leaf, CacheInfo* info) {
   }
 }
 
+typedef struct {
```
gchatelet: Can you add a documentation line?

Issue #108, opened 2020-03-04
Shared build not working?


baentsch:
Positive statement first: All works fine when doing static builds. Our project that wants to use `cpu_features` has to build its libraries shared, though, too, and there I'm running into problems:

I read and followed these suggestions (https://github.com/google/cpu_features/issues/44 ) as well as the [subproject build instructions](https://github.com/google/cpu_features/blob/master/cmake/README.md) but to no avail: 
1) When doing a `cpu_features` "local" `cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DBUILD_PIC=ON -H. -Bcmake_build` this is the result of `make` on Ubuntu:
```[ 77%] Linking C shared library libcpu_features.so
/usr/bin/ld: CMakeFiles/utils.dir/src/string_view.c.o: relocation R_X86_64_PC32 against symbol `CpuFeatures_StringView_IndexOf' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
CMakeFiles/cpu_features.dir/build.make:100: recipe for target 'libcpu_features.so' failed
make[2]: *** [libcpu_features.so] Error 1
CMakeFiles/Makefile2:178: recipe for target 'CMakeFiles/cpu_features.dir/all' failed
make[1]: *** [CMakeFiles/cpu_features.dir/all] Error 2
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
```

2) When doing this embedded in our larger project, not only can I not build the shared lib, even this happens during `make`:
```
[ 95%] Linking C executable list_cpu_features
CMakeFiles/list_cpu_features.dir/src/utils/list_cpu_features.c.o: In function `CreateTree':
list_cpu_features.c:(.text+0x575): undefined reference to `GetX86Info'
list_cpu_features.c:(.text+0x57d): undefined reference to `GetX86CacheInfo'
list_cpu_features.c:(.text+0x585): undefined reference to `FillX86BrandString'
list_cpu_features.c:(.text+0x676): undefined reference to `GetX86Microarchitecture'
list_cpu_features.c:(.text+0x67d): undefined reference to `GetX86MicroarchitectureName'
list_cpu_features.c:(.text+0x6c9): undefined reference to `GetX86FeaturesEnumValue'
list_cpu_features.c:(.text+0x6d9): undefined reference to `GetX86FeaturesEnumName'
[...]

PR #113, opened 2020-03-11
Fix shared build

gadoofou87:
Fixes https://github.com/google/cpu_features/issues/108

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,11 @@
 cmake_minimum_required(VERSION 3.0)
 
+# option() honors normal variables.
+# see: https://cmake.org/cmake/help/git-stage/policy/CMP0077.html
+if(POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ option(BUILD_SHARED_LIBS "Build library as shared." OFF)
 # PIC
 option(BUILD_PIC "Build with Position Independant Code." OFF) # Default is off at least for GCC
 
+if(BUILD_SHARED_LIBS AND UNIX)
```
Mizux: Maybe add a comment:
# Force PIC on unix when building shared libs
# see: https://en.wikipedia.org/wiki/Position-independent_code

Mizux:
also related to #108 we should add [CMP0077](https://cmake.org/cmake/help/git-stage/policy/CMP0077.html)

At the beginning of the file:
```cmake
# option() honors normal variables.
# see: https://cmake.org/cmake/help/git-stage/policy/CMP0077.html
if(POLICY CMP0077)
  cmake_policy(SET CMP0077 NEW)
endif()
```

PR #132, opened 2020-09-22
ci: Add clang-format check

Mizux:


Mizux:
Open question: If a PR is composed of multiple commits does FETCH_HEAD will give us the diff of all or only the last one ?
on my way to test it, I'll keep you posted...

gchatelet:
> Open question: If a PR is composed of multiple commits does FETCH_HEAD will give us the diff of all or only the last one ?
> on my way to test it, I'll keep you posted...

AFAIU it will give the diff for all so we should be safe.

I'm waiting for #134 to be in before merging this one.

.github/workflows/clang_format.yml
```diff
@@ -0,0 +1,20 @@
+name: Clang Format
```
gchatelet, line 1: Maybe `clang-format Check`

Mizux:
after few test in: https://github.com/Mizux/poc_diff/actions
* FETCH_HEAD is null
* you must use git fetch since the "git clone" in the job use depth=1
* you can then use `git diff --name-only origin/master`

.github/workflows/clang_format.yml
Mizux: done

gchatelet:
> after few test in: https://github.com/Mizux/poc_diff/actions
> 
> * FETCH_HEAD is empty e.g. https://github.com/Mizux/poc_diff/runs/1153735002?check_suite_focus=true
> * you must use git fetch since the "git clone" in the job use depth=1
> * you can then use `git diff --name-only origin/master`
>   e.g. This PR of two commits [Mizux/poc_diff#1](https://github.com/Mizux/poc_diff/pull/1)
>   all files are correctly listed: https://github.com/Mizux/poc_diff/pull/1/checks?check_run_id=1153803609

In case of big repository it may be worth limiting the fetch to the last 50 commits, see [`git fetch --depth`](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---depthltdepthgt)

Mizux:
ok I'll test first on my poc repo

gchatelet:
> ok I'll test first on my poc repo

Actually [`--negotiation-tip`](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---negotiation-tipltcommitglobgt) seems like a better option

Mizux:
```sh
Run git fetch --negotiation-tip=origin/master
fatal: origin/master is not a valid object
##[error]Process completed with exit code 128.
```
ref: https://github.com/Mizux/poc_diff/pull/1/checks?check_run_id=1153860263

I'll continue to dig...

gchatelet:
Ok let's use `git fetch` for now since this repo is quite small. It would be nice to have for other bigger repo though but we can do it as a separate step.
If you agree I'll `Squash and merge` the PR as-is.

Mizux:
￼So I've used `git fetch origin master` which will set `FETCH_HEAD` so we can use it...

gchatelet:
Thx a lot @Mizux :+1:

PR #137, opened 2020-10-10
[NFC] Generate separate tables via macro

gchatelet:
This is a non functional change, it allows:
 - Getting rid of `unix_features_aggregator`
 - Have a single blob describing the features
 - Fix wrong mocking of `hwcaps`

Downside: abuse of macros makes the code slightly magical and harder to understand.
It think it's still an improvement over the current situation as there's less repetition and less chances to get something wrong.


gchatelet:
@Mizux `clang-format Check` is unhappy with deleted files :-/

Mizux:
> @Mizux `clang-format Check` is unhappy with deleted files :-/

We should use something like this in the job
```sh
git diff --diff-filter=ACMRTUXB
```
i.e. filter out D aka deleted files

man git:
```
       --diff-filter=[(A|C|D|M|R|T|U|X|B)...[*]]
           Select only files that are Added (A), Copied (C), Deleted (D), Modified (M), Renamed (R), have their type
           (i.e. regular file, symlink, submodule, ...) changed (T), are Unmerged (U), are Unknown (X), or have had
           their pairing Broken (B). Any combination of the filter characters (including none) can be used. When *
           (All-or-none) is added to the combination, all paths are selected if there is any file that matches other
           criteria in the comparison; if there is no file that matches other criteria, nothing is selected.

           Also, these upper-case letters can be downcased to exclude. E.g.  --diff-filter=ad excludes added and
           deleted paths.

           Note that not all diffs can feature all types. For instance, diffs from the index to the working tree can
           never have Added entries (because the set of paths included in the diff is limited by what is in the
           index). Similarly, copied and renamed entries cannot appear if detection for those types is disabled.
```

src/define_tables.h
```diff
@@ -0,0 +1,72 @@
+// Copyright 2020 Google Inc.
```
Mizux, line 1: now it's Google LLC IIRC

gchatelet: Thx I'll fix them for the whole code base in a separate patch then update only this one here.

Mizux (approved):

PR #142, opened 2020-10-16
Provides a release script

### LINKED ISSUE #141: CMake: version management
Currently the version is hard coded directly in `CMakeLists.txt`, e.g.:
https://github.com/google/cpu_features/blob/a8397ba4591237c17d18e4acc091f5f3ebe7391e/CMakeLists.txt#L9

So each time a new release is pushed, using the github release interface, you must do the sync manually with the CMakeLists.txt

We should instead try to generate it from the last git tag (and fallback to 9.9.9 or 9999 like gentoo) IMHO

## Proposal
A small proposal (not tested!!!)

CMakeLists.txt:
```cmake
list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
include(utils)
set_version(VERSION)
project(CpuFeatures VERSION ${VERSION} LANGUAGES C)
```
cmake/utils.cmake:
```cmake
function(get_version_from_git_tag GIT_VERSION)
  find_package(Git QUIET)
  if(NOT GIT_FOUND)
    message(STATUS "Did not find git package.")
    set(VERSION 9.9.9)
  else()
    # If no tags can be found, it is a git shallow clone
    execute_process(COMMAND
      ${GIT_EXECUTABLE} describe --tags
      RESULT_VARIABLE _OUTPUT_VAR
      OUTPUT_VARIABLE FULL
      ERROR_QUIET
      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
    if(NOT _OUTPUT_VAR)
      set(VERSION "${RESULT_VARIABLE}")
    else()
      message(STATUS "Did not find any tag.")
      set(VERSION "9.9.9")
    endif()
  endif()
  set(${GIT_VERSION} "${VERSION}" PARENT_SCOPE)
endfunction()

function(set_version VERSION)
  # Retrieve version from git tag if .git is present otherwise set it to 9.9.9
  get_filename_component(GIT_DIR ".git" ABSOLUTE)
  if(EXISTS ${GIT_DIR})
    get_version_from_git_tag(GIT_VERSION)
  else()
    # If not a git repo force version to 9.9.9
    set(GIT_VERSION "9.9.9")
  endif()
  set(${VERSION} "${GIT_VERSION}" PARENT_SCOPE)
endfunction()
```

ref: Gentoo 9999 ebuild https://forums.gentoo.org/viewtopic-p-7428490.html

gchatelet:
Fixes #141

### DIFF
--- a/scripts/make_release.sh
+++ b/scripts/make_release.sh
@@ -66,7 +66,7 @@ echo -e "${ACTION}Commit new revision${NOCOLOR}"
 git add CMakeLists.txt
 git commit -m"Release ${GIT_TAG}"
 
-echo -e "${ACTION}Commit new revision${NOCOLOR}"
+echo -e "${ACTION}Create new tag${NOCOLOR}"
 git tag ${GIT_TAG}
 
 echo -e "${FINISHED}Local release is ready. Run `git push origin --tags`${NOCOLOR}"
--- a/scripts/make_release.sh
+++ b/scripts/make_release.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+set -e # Fail on error
+
+ACTION='\033[1;90m'
+FINISHED='\033[1;96m'
+READY='\033[1;92m'
+NOCOLOR='\033[0m'
+ERROR='\033[0;31m'
+
+echo -e "${ACTION}Checking environnement${NOCOLOR}"
+if [[ ! $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
+then
+    echo -e "${ERROR}Invalid version number. Aborting. ${NOCOLOR}"
+    exit 1
+fi
+
+declare -r VERSION=$1
+declare -r GIT_TAG="v$1"
+
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+if [[ "${BRANCH}" != "master" ]]
+then
+    echo -e "${ERROR}Not on master. Aborting. ${NOCOLOR}"
+    echo
+    exit 1
+fi
+
+git fetch
+HEADHASH=$(git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


scripts/make_release.sh
```diff
@@ -0,0 +1,72 @@
...
+set -e # Fail on error
+
+ACTION='\033[1;90m'
+FINISHED='\033[1;96m'
+READY='\033[1;92m'
+NOCOLOR='\033[0m'
+ERROR='\033[0;31m'
+
+echo -e "${ACTION}Checking environnement${NOCOLOR}"
+if [[ ! $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
```
Mizux, line 12: should we use full sem ver regex ? (overkill IMHO)
https://semver.org/ (bottom)
https://regex101.com/r/Ly7O1x/3/

note: CMake also allow a "tweak" field
https://cmake.org/cmake/help/latest/command/project.html#options
note: I've never use it, but sometine it is use when you just want to bump a dependency/compiler config but your source code is not touched, not sure it is releavant for cpu_features

scripts/make_release.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
Mizux, line 1: note: #!/usr/bin/env bash
is more portable IMHO
ref: https://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html

Mizux (approved):

scripts/make_release.sh
gchatelet: Let's keep it simple for now and improve if we feel it's needed.

Mizux (approved):

PR #154, opened 2021-04-06
Fix clang-format not being installed (fixes build error on #153)

natanbc:
clang-format was moved to the `clang-extra-tools` package on current alpine edge


Mizux:
Sorry, I've just merged the duplicate PR #155

PR #156, opened 2021-04-12
Fix C++ namespace in README

guillaumekln:
The correct namespace is `cpu_features` as defined in: https://github.com/google/cpu_features/blob/3cc8f310d950b44cc6d2ead8057cf43330a726fd/include/cpu_features_macros.h#L108


Mizux (approved):

Mizux:
thx !

PR #159, opened 2021-06-09
Support Freebsd

mldulaney:
This patch introduces support for x86 on FreeBSD.  I note that I added the check for FreeBSD on aarch64 to fail in case this gets built on an aarch64 FreeBSD system.


Mizux (approved):

vishwin:
Those sysctl names are probably wrong. On a Broadwell system, only `hw.instruction_sse` is present, yet the hardware actually supports much more. `cpuinfo` from `linprocfs` is correct, however, but FreeBSD binaries should not be reading from any Linuxulator interfaces.

mldulaney:
Hi, any chance this could get merged?

src/cpuinfo_x86.c
```diff
@@ -1244,6 +1246,13 @@ static void DetectSseViaOs(X86Features* features) {
   features->ssse3 = GetDarwinSysCtlByName("hw.optional.supplementalsse3");
   features->sse4_1 = GetDarwinSysCtlByName("hw.optional.sse4_1");
   features->sse4_2 = GetDarwinSysCtlByName("hw.optional.sse4_2");
+#elif defined(CPU_FEATURES_OS_FREEBSD)
+  features->sse = sysctlbyname("hw.instruction_sse", NULL, NULL, NULL, 1);
```
gchatelet, line 1250: As far as I understand, this is not how `sysctlbyname` is supposed to be used.
https://www.freebsd.org/cgi/man.cgi?query=sysctlbyname&manpath=FreeBSD+13.0-RELEASE+and+Ports

One need to set proper arguments.
Also returned value is `0` upon success and `-1` upon failures.
I believe we should rename `GetDarwinSysCtlByName` above and use it here.

src/cpuinfo_aarch64.c
```diff
@@ -80,6 +80,11 @@
 #define DEFINE_TABLE_FEATURE_TYPE Aarch64Features
 #include "define_tables.h"
 
+#if defined(CPU_FEATURES_OS_FREEBSD)
+#error "FreeBSD not yet supported on this arch"
+#endif  // CPU_FEATURES_OS
```
gchatelet, line 85: `// CPU_FEATURES_OS_FREEBSD`

src/cpuinfo_x86.c
```diff
@@ -107,6 +107,8 @@
 #error "Darwin needs support for sysctlbyname"
 #endif
 #include <sys/sysctl.h>
+#elif defined(CPU_FEATURES_OS_FREEBSD)
```
gchatelet, line 110: This should probably be merged with the Darwin logic above.

gchatelet (changes requested):

gchatelet:
I'm closing this PR, FreeBSD support will be handled by #163 which will be submitted as soon as we have FreeBSD support in our CI.

Thank you @mldulaney for the patch: it didn't make it to the repo but it got me interested in solving the issue so it's a worthwhile contribution! (also thank you @vishwin for the comment)

PR #161, opened 2021-06-22
[NFC] refactor the code so it's easier to understand the execution flow

gchatelet:
This is a non functional change, it's a small refactoring with more comments so it's easier to follow along.

### DIFF
--- a/src/cpuinfo_x86.c
+++ b/src/cpuinfo_x86.c
@@ -1181,41 +1181,11 @@ static bool GetDarwinSysCtlByName(const char* name) {
 // Internal structure to hold the OS support for vector operations.
 // Avoid to recompute them since each call to cpuid is ~100 cycles.
 typedef struct {
-  bool have_sse_via_os;
-  bool have_sse_via_cpuid;
+  bool have_sse;
   bool have_avx;
   bool have_avx512;
   bool have_amx;
 } OsSupport;
-
-static const OsSupport kEmptyOsSupport;
-
-static OsSupport CheckOsSupport(const uint32_t max_cpuid_leaf) {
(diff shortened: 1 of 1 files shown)


src/cpuinfo_x86.c
```diff
@@ -1376,7 +1373,7 @@ static void ParseExtraAMDCpuId(X86Info* info, OsSupport os_support) {
 
   X86Features* const features = &info->features;
 
-  if (os_support.have_sse_via_cpuid) {
+  if (os_support.have_sse) {
```
Mizux, line 1376: are you sure of this ? I mean on an older atom you won't have sse access through cpuid (thus the previous variable)
EDIT: ok specific code path for AMD processor, -> do you know if AMD have built some processors with the same issue than the older ATOM (aka no cpuid support of sse) ?

Mizux (approved):

src/cpuinfo_x86.c
gchatelet: So you can always query x86 cpus for supported vector instructions via `cpuid`. But you can't always know whether the OS will save/restore these register during context switches via `cpuid` (sometime you need to call OS specific functions).

Here `os_support.have_sse` means we know OS will save/restore sse registers and we ask whether the CPU supports the `sse4a` AMD instructions.

That said I realize that the `have_` part is confusing. I will just use `os_support.sse` or find something more explicit as the semantic is not super clear. I'll update the patch.

gchatelet: I think it's clearer with https://github.com/google/cpu_features/pull/161/commits/d818652dfbd89fe8ded2b2db58f1bf1420db6441. Let me know what you think.

Mizux (approved):
LGTM

PR #162, opened 2021-06-23
fix: Return default value from ‘GetCacheTypeString’

kriskwiatkowski:
The build fails with following message when -Werror
and -Werror=return-type are enabled.

In function ‘GetCacheTypeString’:
	error: control reaches end of non-void function [-Werror=return-type]

Simple fix is to return 0 from that function. The
return value is then checked by an assert in the
AddMapEntry. The assert will only work in debug mode,
but that seems to be enough.

### DIFF
--- a/include/cpu_features_macros.h
+++ b/include/cpu_features_macros.h
@@ -213,4 +213,17 @@
 #endif  //  defined(__mips_msa)
 #endif  //  defined(CPU_FEATURES_ARCH_MIPS)
 
+////////////////////////////////////////////////////////////////////////////////
+// Utils
+////////////////////////////////////////////////////////////////////////////////
+
+// Communicates to the compiler that the block is unreachable
+#if defined(CPU_FEATURES_COMPILER_CLANG) || defined(CPU_FEATURES_COMPILER_GCC)
+#define UNREACHABLE() __builtin_unreachable()
+#elif defined(CPU_FEATURES_COMPILER_MSC)
+#define UNREACHABLE() __assume(0)
+#else
+#define UNREACHABLE()
+#endif
+
 #endif  // CPU_FEATURES_INCLUDE_CPU_FEATURES_MACROS_H_
--- a/src/utils/list_cpu_features.c
+++ b/src/utils/list_cpu_features.c
@@ -340,6 +340,7 @@ static Node* GetCacheTypeString(CacheType cache_type) {
     case CPU_FEATURE_CACHE_PREFETCH:
       return CreateConstantString("prefetch");
   }
+  UNREACHABLE();
 }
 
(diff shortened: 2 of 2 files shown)


src/utils/list_cpu_features.c
```diff
@@ -340,6 +340,8 @@ static Node* GetCacheTypeString(CacheType cache_type) {
     case CPU_FEATURE_CACHE_PREFETCH:
       return CreateConstantString("prefetch");
   }
+  /* Should never happen */
+  return 0;
```
Mizux, line 344: I would use `NULL`
ref: https://en.cppreference.com/w/c/types/NULL

src/utils/list_cpu_features.c
```diff
@@ -340,6 +340,8 @@ static Node* GetCacheTypeString(CacheType cache_type) {
     case CPU_FEATURE_CACHE_PREFETCH:
       return CreateConstantString("prefetch");
   }
+  /* Should never happen */
```
gchatelet, line 343: Since this should never happen I'd rather use `__builtin_unreachable()` for clang/gcc and `__assume(0)` for MSVC.
Something along these lines:
```
#if defined(CPU_FEATURES_COMPILER_CLANG) || defined(CPU_FEATURES_COMPILER_GCC)
  __builtin_unreachable();
#elif defined(CPU_FEATURES_COMPILER_MSC)
  __assume(0);
#else
#error "Unsupported compiler."
#endif
```

kriskwiatkowski: ok, done. thanks, I wasn't aware of those builtins.
Instead of issuing an error for compiler other than MSC,GCC,clang I simply do nothing. In this unusual situation, it is a bit better as compilation may still work if something like ``-pedantic`` is not used.

include/cpu_features_macros.h
```diff
@@ -213,4 +213,16 @@
 #endif  //  defined(__mips_msa)
 #endif  //  defined(CPU_FEATURES_ARCH_MIPS)
 
+////////////////////////////////////////////////////////////////////////////////
+// Utils
+////////////////////////////////////////////////////////////////////////////////
+// Communicates to the compiler that the block is unreachable
```
gchatelet, line 219: Add a new line please

gchatelet (commented):
Thx! I was about to suggest to add this to `include/cpu_features_macros.h`, you read my mind : )
One nit on a missing new line and this is good to go!

include/cpu_features_macros.h
kriskwiatkowski: ah, right. done!

kriskwiatkowski:
Thanks!

PR #166, opened 2021-06-29
Add GitHub workflows

Mizux:


Mizux:
@gchatelet On my way to fix Travis jobs -> please wait

CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ endif(NOT CMAKE_BUILD_TYPE)
 
 # BUILD_TESTING is a standard CMake variable, but we declare it here to make it
 # prominent in the GUI.
-option(BUILD_TESTING "Enable test (depends on googletest)." OFF)
+option(BUILD_TESTING "Enable test (depends on googletest)." ON)
```
gchatelet, line 22: I'd rather have it set on the cmake command line as it triggers an additional dependency.
It seems it's already ON for VS below (`-DBUILD_TESTING=ON`)

gchatelet (changes requested):

CMakeLists.txt
Mizux: Actually `include(CTest)` automatically set it to ON, so I wanted to use the default value (most distro usually have cmake wrapper/alias which will set it to OFF "by default")

gchatelet: From https://cmake.org/cmake/help/latest/module/CTest.html
`The module automatically creates a BUILD_TESTING option that selects whether to enable testing support (ON by default)`

So do we even need to declare it? (since we `include(CTest)` later in the file)

gchatelet: So as discussed, let's split this patch in two:
 - First remove the `BUILD_TESTING` option (since we `include(CTest)`)
 - Then, rebase this patch and submit with the default behaviour (i.e. `BUILD_TESTING=ON`)

PR #177, opened 2021-07-08
CMake: add BUNDLE DESTINATION to fix cross-build to iOS/watchOS/tvOS 

SpaceIm:
BUNDLE DESTINATION is required for installation of executable target if cross-build to iOS/tvOS/watchOS

see https://cmake.org/cmake/help/latest/policy/CMP0006.html


SpaceIm:
@googlebot I signed it!

prince-chrismc (approved):

PR #186, opened 2021-10-24
ARM64 detection via asm 

toor1245:
Added an initial implementation, detecting ARM64 through assembly instructions, thereby getting rid of OS binding, which can be used, for example, to detect ARM on Windows.

What do you think?


include/internal/bit_utils.h
```diff
@@ -27,8 +27,8 @@ inline static bool IsBitSet(uint32_t reg, uint32_t bit) {
   return (reg >> bit) & 0x1;
 }
 
-inline static uint32_t ExtractBitRange(uint32_t reg, uint32_t msb,
-                                       uint32_t lsb) {
+inline static uint64_t ExtractBitRange(uint64_t reg, uint64_t msb,
+                                       uint64_t lsb) {
```
toor1245, line 31: fixed overflow when occurs extract bit range from ID_AA64_*

toor1245:
@gchatelet, @Mizux this pull request is ready for review.

In generally I tested it on Linux with proc/cpuinfo and new implementation through inline asm and checked behaviour with fully implementation in case only inline asm:

_Linux proc/cpuinfo + MIDR_
image

_Linux DetectFeaturesBase + MIDR_
image

As you can see detection features the same, except  evtstrm, since HWCAP_EVTSTRM is generic timer is configured to generate events at a frequency of approximately 100KHz, and we cannot get this info through cpuid like everyone else.

All other cpu features which we have in `Aarch64Features` struct was added. I think this should work for Unix, but not Windows yet. Implementation wasn't added, since msvc compiler on Aarch64 __asm keyword  is not supported, then I tried using `armasm64` this is the same `armasm` but microsoft with some diference [microsoft armasm directives](https://docs.microsoft.com/bs-cyrl-ba/cpp/assembler/arm/arm-assembler-reference?view=msvc-140)
in GNU assembler is used [`.inst`](https://developer.arm.com/documentation/100068/0608/migrating-from-armasm-to-the-armclang-integrated-assembler/data-definition-directives) directive, and for armasm `DCI`, but it doesn't work, it is an unknown directive.

toor1245:
could you provide dev/aarch64 branch, where we can add CI/CD and if there are defects to fix them, improve tests and only then merge it into the master or is it unnecessary?

toor1245:
image
checked via qemu and it works on freebsd

arkivm:
@toor1245 does it work on aarch64-darwin by any chance? I tried running it, it hits an "illegal instruction".

toor1245:
> @toor1245 does it work on aarch64-darwin by any chance? I tried running it, it hits an "illegal instruction".

Hi, thanks for the reply! I think no, for now. Right now I have no opportunity to test, and there is no CI to test yet. An issue has been created to add CI jobs for ios, https://github.com/google/cpu_features/issues/197, as a temporary solution can be done through sysctlbyname, but I would not like this since there will be a dependence on the OS, and lowers the possibility for further improving the information of a given architecture

arkivm:
Btw, if this gets merged, are we going to drop all the OS APIs (e.g., `sysctl`) for arm64 and switch to this version?

toor1245:
> Btw, if this gets merged, are we going to drop all the OS APIs (e.g., `sysctl`) for arm64 and switch to this version?

for Linux / Adnroid, something that has been working for a long time and does not break the public api, I think definitely not now, but on other OS this api may become truncated, what's not good

toor1245:
Since a new pull request was introduced, I converted this pull request to draft until it was merged https://github.com/google/cpu_features/pull/204 to the master branch

gchatelet:
Thx for the work here. I'm keeping an eye on the PR (even though I haven't commented lately). Just to make sure we're on the same page, there should be only one header for aarch64 eventually. I assume you have the two headers in parallel so to check that the results are consistent?

toor1245:
> Thx for the work here. I'm keeping an eye on the PR (even though I haven't commented lately). Just to make sure we're on the same page, there should be only one header for aarch64 eventually. I assume you have the two headers in parallel so to check that the results are consistent?

@gchatelet at the moment, this PR has only one `include/cpuinfo_aarch64.h` header unchanged and added two headers to `include/internal` as `include/internal/cpuid_x86.h`, or do we need to move the implementation? Sorry, but I misunderstood the question of the consistency of the two headers.

I think we сan to create small pull-requests, since PR too big and hard to review, and leave this PR as a sample.

I see that we can create at least:
1. Add new inl file `stringize.inl`, since often operantion and dependency of `define_introspection.inl` .
2. Fix or add override function of ExctractBitRange uint64_t, as there is an overflow when extracting 64 bit info in aarch64.
2. Add `include/internal/cpuid_aarch64.h`, `src/define_cpuid_aarch64.inl` and `impl_aarch64__base_implementation.inl` where only MIDR_EL1(architecture, revision, part num, variant) will be implemented.
3. Provide new inline asm testing logic.
4. Provide PR for each ID_* separately.

And as I mentioned above, it will be good if we create a dev branch for implementation.

 What do you think?

toor1245:
### Update
I have completed research on aarch64 support for Windows. Access to system registers is limited from EL0 and we can't get information from user mode, to solve this problem we have only one way, we have to write software driver in kernel mode, where we have more access, and make client communicate with driver.

_here is the result of the test driver of load and unload_
Screenshot (8)

to read processor info we can use `_ReadStatusReg`, thus we don't have to think about armasm64 integration  
https://reviews.llvm.org/D53115
https://docs.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics?view=msvc-170

_in lib we have to add something like this:_
```C
HANDLE hDevice = CreateFile(L"\\\\.\\CpuFeatures", GENERIC_READ,
          0, NULL, OPEN_EXISTING, 0, NULL);

 if (hDevice == INVALID_HANDLE_VALUE)
      // handle hDevice == INVALID_HANDLE_VALUE

 EL1Info info;
 DWORD bytes;
 BOOL ok = ReadFile(hDevice, &info, sizeof(EL1Info), &bytes, NULL);
 if (!ok)
     // handle !ok
 if (bytes != sizeof(EL1Info))
     // handle bytes != sizeof(EL1Info)
 CloseHandle(hDevice);
```

Linux allows reading information from EL1 for the following system registers:
* MIDR_EL1
* ID_AA64ISAR0_EL1
* ID_AA64PFR0_EL1
* ID_AA64PFR1_EL1
* ID_AA64ISAR1_EL1
* ID_AA64MMFR0_EL1
* ID_AA64MMFR2_EL1
* ID_AA64ZFR0_EL1
* ID_AA64MMFR1_EL1
* ID_AA64ISAR2_EL1
https://www.kernel.org/doc/html/latest/arm64/cpu-feature-registers.html 

@gchatelet, @Mizux, what do you think?

toor1245:
hi @arkivm, could you please run this code on your M1 to make sure the behavior with EL on macOS is the same as it is on Windows

compiler clang.
https://reviews.llvm.org/D53115

```C
#include <stdint.h>
#include <intrin.h>

#define ARM64_SYSREG(op0, op1, crn, crm, op2) \
        ( ((op0 & 1) << 14) | \
          ((op1 & 7) << 11) | \
          ((crn & 15) << 7) | \
          ((crm & 15) << 3) | \
          ((op2 & 7) << 0) )

#define ARM64_CNTVCT            ARM64_SYSREG(3, 3, 14, 0, 2)

int main() {
   uint64_t cntvct = _ReadStatusReg(ARM64_CNTVCT); // should work
   return 0;
}
```

```C
#include <stdint.h>
#include <intrin.h>

#define ARM64_SYSREG(op0, op1, crn, crm, op2) \
        ( ((op0 & 1) << 14) | \
          ((op1 & 7) << 11) | \
          ((crn & 15) << 7) | \
          ((crm & 15) << 3) | \
          ((op2 & 7) << 0) )

#define ARM64_MIDR_EL1         ARM64_SYSREG(3, 0, 0, 0, 0)

int main() {
   uint64_t midr = _ReadStatusReg(ARM64_MIDR_EL1); // the probability that it will fail the way it is done in user mode
   return 0;
}
```

arkivm:
@toor1245

```c
__asm__ volatile ("mrs %0, cntvct_el0" : "=r" (cntvct1)); // works
__asm__ volatile ("mrs %0, tpidrro_el0" : "=r" (tpidr));  // works
__asm__ volatile ("mrs %0, midr_el1" : "=r" (midr)); // fails - illegal instruction
```

I couldn't get the`_ReadStatusReg` intrinsic to work. When I add `intrin.h`, it was complaining
```c
include/intrin.h:12:15: fatal error: 'intrin.h' file not found
#include_next <intrin.h>
              ^~~~~~~~~~
```

Let me know if you want me to test something else.

toor1245:
@arkivm, thanks for the help!

PR #198, opened 2021-10-28
API change: embed `brand_string` in GetX86Info()

gchatelet:
This is an API change.
It reduces the chance that user passes an invalid buffer to `FillBrandString`


toor1245:
@gchatelet, do we need to close this PR? Since already merged into master https://github.com/google/cpu_features/pull/214

PR #199, opened 2021-11-12
Add cache detection for old AMD processors

toor1245:
In this pull request I added cache info detection for old AMD processors. We can get information about the TLB, but I don't know how to process this information as there is a difference in leaves, there is information for 4KB, 2MB, 4MB, 1GB pages, so I have defined these fields as UNDEF so far

### DIFF
--- a/src/impl_x86__base_implementation.inl
+++ b/src/impl_x86__base_implementation.inl
@@ -1757,6 +1757,8 @@ static CacheLevelInfo GetCacheL3InfoLegacyAMD(const uint32_t cache_id,
   info->size = 4;
 
   const int ways = GetWaysLegacyAMD(cache_id);
+  // https://www.amd.com/system/files/TechDocs/25481.pdf page 25
+  // CPUID Fn8000_0006_EDX L3 Cache Identifiers
   return GetCacheLevelInfoLegacyAMD(cache_id, cache_type, cache_size * 512, 3,
                                     ways);
 }
--- a/src/impl_x86__base_implementation.inl
+++ b/src/impl_x86__base_implementation.inl
@@ -1686,6 +1686,104 @@ static void ParseCacheInfo(const int max_cpuid_leaf, uint32_t leaf_id,
   if (info.size > 0) *old_info = info;
 }
 
+static CacheLevelInfo GetCacheLevelInfoLegacyAMD(uint32_t cache_id,
+                                                 CacheType cache_type,
+                                                 int cache_size, int level,
+                                                 int ways) {
+  const int KiB = 1024;
+  const int UNDEF = -1;
+  return (CacheLevelInfo){.level = level,
+                          .cache_type = cache_type,
+                          .cache_size = cache_size * KiB,
+                          .ways = ways,
+                          .line_size = ExtractBitRange(cache_id, 7, 0),
+                          .tlb_entries = UNDEF,
+                          .partitioning = UNDEF};
+}
+
+// https://www.amd.com/system/files/TechDocs/25481.pdf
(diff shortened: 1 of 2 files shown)


toor1245:
@gchatelet, @Mizux, please review

toor1245:
rebased, @gchatelet, @Mizux, please review PR

toor1245:
@gchatelet, this PR has been here for a long time, please review or is this inappropriate PR and should we close it?

src/impl_x86__base_implementation.inl
```diff
@@ -1620,6 +1620,104 @@ static void ParseCacheInfo(const int max_cpuid_leaf, uint32_t leaf_id,
...
+  const int ways = GetWaysLegacyAMD(cache_id);
+  return GetCacheLevelInfoLegacyAMD(cache_id, cache_type, cache_size, 2, ways);
+}
+
+static CacheLevelInfo GetCacheL3InfoLegacyAMD(const uint32_t cache_id,
+                                              CacheType cache_type,
+                                              CacheInfo* info) {
+  const uint32_t cache_size = ExtractBitRange(cache_id, 31, 18);
+  if (cache_size == 0) {
+    info->size = 3;
```
gchatelet, line 1688: If no L3 then info->size should be 2 right?

src/impl_x86__base_implementation.inl
```diff
@@ -1620,6 +1620,104 @@ static void ParseCacheInfo(const int max_cpuid_leaf, uint32_t leaf_id,
...
+    return kEmptyCacheLevelInfo;
+  }
+  info->size = 4;
```
gchatelet, line 1691: And this should be 3

src/impl_x86__base_implementation.inl
```diff
@@ -1620,6 +1620,104 @@ static void ParseCacheInfo(const int max_cpuid_leaf, uint32_t leaf_id,
...
+
+  const int ways = GetWaysLegacyAMD(cache_id);
+  return GetCacheLevelInfoLegacyAMD(cache_id, cache_type, cache_size * 512, 3,
```
gchatelet, line 1694: Can you add the source documentation for `cache_size * 512`?

src/impl_x86__base_implementation.inl
toor1245: According to AMD documentation, we have L1 data Fn8000_0005_ECX and instruction Fn8000_0005_EDX cache, for L2 and L3 we have unified via Fn8000_0006_ECX and Fn8000_0006_EDX, thus it should be 3 if no L3 (L1D, L1I, L2)

src/impl_x86__base_implementation.inl
toor1245: done

src/impl_x86__base_implementation.inl
toor1245: answer above

src/impl_x86__base_implementation.inl
gchatelet: I see, why not start with `size==0` and do `++(info->size)` in each of the function adding a cache level entry?
It seems to me that it's easier to understand this way. WDYT?

toor1245: good point, I fully agree with you, it's also easier for me to understand, will update with pass of `int* cache_level` instead of `CacheInfo*`, thanks!

toor1245: @gchatelet, I updated PR. I got rid of the duplicate logic a bit and introduced an internal structure to create the `CacheLevelInfo` and increment `info->size` in the loop

gchatelet:
Looks much better thx!

PR #202, opened 2021-11-16
Fix: CMake googletest-download failed

toor1245:


toor1245:
@Mizux, @gchatelet please review

Mizux:
Thx !

Maybe we should point to a release instead like `v1.11.0`
pro: more stable.
cons: increase the technical debt and maintenance cost to bump it from time to time to try to stay up to date.

PR #204, opened 2021-11-18
Add support for Apple M1 AArch64 SoCs

arkivm:
Completely based on #150. Thanks to @sbehnke!
+ Refactoring to accomodate the new source tree
+ Adding more feature flags

As #150 is stale, I am just trying to push it across the finish line.

PR #150, opened 2020-12-23
Adding support for Apple M1 AArch64 Architecture Processor.

sbehnke:
This is the initial attempt at adding support for the new Apple Silicon M1 cpu to cpu_features. Based on my own laptop I tried to make my best guess for how the ARM features map from the `sysctl` optional values to the `AArch64Features`.

```
hw.optional.floatingpoint: 1
hw.optional.watchpoint: 4
hw.optional.breakpoint: 6
hw.optional.neon: 1
hw.optional.neon_hpfp: 1
hw.optional.neon_fp16: 1
hw.optional.armv8_1_atomics: 1
hw.optional.armv8_crc32: 1
hw.optional.armv8_2_fhm: 1
hw.optional.armv8_2_sha512: 1
hw.optional.armv8_2_sha3: 1
hw.optional.amx_version: 2
hw.optional.ucnormal_mem: 1
hw.optional.arm64: 1
```

I also tried to map between the variant, revision, part and implementor based on the following sysctl values. I'm not sure if I got them correct, but it is a good first stab at it.

```
hw.cputype: 16777228
hw.cpusubtype: 2
hw.cpu64bit_capable: 1
hw.cpufamily: 458787763
hw.cpusubfamily: 2
```

The output I get is:

```
./list_cpu_features
arch            : aarch64
implementer     : 16777228 (0x100000C)
variant         :   2 (0x02)
part            : 458787763 (0x1B588BB3)
revision        :   2 (0x02)
flags           : asimdfhm,atomics,crc32,fp,fphp,sha3,sha512
```

This pull request is an effort to address:
https://github.com/google/cpu_features/issues/121
https://github.com/gnuradio/volk/issues/428

PR #208, opened 2021-11-19
New code layout for list_cpu_features

toor1245:
This pull request is intended for architecture isolation and further OS implementation. @gchatelet, what do you think about it?


angusryer:
@gchatelet @toor1245 -- So it seems that the code layout changes referenced in #150 #186 #194 #204 #208 #209 #210 #211 (which mostly relate to supporting arm64) have been split between #194 (which was merged already) and this PR, is that correct?

And that the main PR for adding support for the M1 is #204, but it is too large for review, correct?

And finally, that the layout changes themselves seem to be the major blocker for the PRs related to supporting the M1 (most of the PRs listed above), is that also correct?

Forgive my ignorance if that's totally off--I'm just trying to understand the situation.

toor1245:
@angusryer, This PR is not related to #194, it's just a suggestion for the code health of list_cpu_features, and doesn't block M1 in any way, since list_cpu_features just uses the cpu_features API for demo

PR #150 is deprecated. 
PR #186 aarch64 detection works via inline assembly for FreeBSD and Linux, for Windows and macOs I'm investigating for now.
PR #204 is based on #150 plus works with new code layout #194 and detects more cpu flags.

angusryer:
Thanks @toor1245 for the clarification. I'll follow 186 and 204 more closely since these seem to be pointing toward a fix for the compile error I'm getting on my M1 for a package that has cpu_features as its dependency.

PR #212, opened 2021-12-28
Add documentation on current behavior for X86

toor1245:


toor1245:
@gchatelet, @Mizux, please review PR

src/impl_x86__base_implementation.inl
```diff
@@ -121,7 +121,22 @@ static Leaves ReadLeaves() {
...
+//  2. Check that the SIMD extensions supports by hardware.
+//     Detect CPUID.1:ECX.XSAVE[bit 26]
+//  3. Check enabled state in XCR0 via XGETBV.
+//  4. Check feature flag for instruction set.
+//
+//  Note:
+//  It is unwise for an application to rely exclusively on feature flag for
+//  instruction set or at all on CPUID.1:ECX.XSAVE[bit 26]: These indicate
+//  hardware support but not operating system support. If XMM/YMM/ZMM/TMM state
+//  management is not enabled by an operating systems, instructions will #UD
```
gchatelet, line 139: AFAIU this is documenting the current behavior right?
If so I would phrase it differently because it reads as if it was an advice to the user.

Also the section to which you added the documentation only contains the following functions:
```
static bool HasMask(uint32_t value, uint32_t mask);
static bool HasXmmOsXSave(uint32_t xcr0_eax);
static bool HasYmmOsXSave(uint32_t xcr0_eax);
static bool HasZmmOsXSave(uint32_t xcr0_eax);
static bool HasTmmOsXSave(uint32_t xcr0_eax);
```

I agree that we could use some documentation to describe what it means to support SIMD instructions but then I would put this section at the top of the file and describe it in terms of process, not in the form of an advice ("it should check", "It is unwise"). Something along these lines:
```
A note on x86 SIMD instructions availability
--------------------------------------------
A number of conditions need to be met for an application to use SIMD instructions:
1. The CPU itself must support the instruction.
    - we use `CPUID` to check whether the feature is supported.
2. The OS must save and restore the associated SIMD register across context switches, we check that:
    - the CPU reports supporting hardware context switching instructions via CPUID.1:ECX.XSAVE[bit 26]
    - the OS reports supporting hardware context switching instructions via CPUID.1:ECX.OSXSAVE[bit 27]
    - the CPU extended control register 0 (XCR0) is set to save and restore the needed SIMD registers

Note that if `XSAVE`/`OSXSAVE` are missing, we delegate the detection to the OS via the `DetectFeaturesFromOs` function or via microarchitecture heuristics.
```

toor1245:
@gchatelet, I updated PR, but there is a question related to `GetX86CacheInfo`, I see that the comment describes that it works only on Intel CPUs, but it works on AMD (new) and Hygon processors. May I update this information or did you do it on purpose?

```C
// Returns cache hierarchy informations.
// Can call cpuid multiple times.
// Only works on Intel CPU at the moment.
CacheInfo GetX86CacheInfo(void);
```

gchatelet:
@toor1245 the comment is outdated, yes please remove the part about "Intel only"

src/impl_x86__base_implementation.inl
```diff
@@ -13,6 +13,65 @@
...
+//   needed SIMD registers
+//
+// Note that if `XSAVE`/`OSXSAVE` are missing, we delegate the detection to the
+// OS via the `DetectFeaturesFromOs` function or via microarchitecture
+// heuristics.
+//
+// Encoding
+// -----------------------------------------------------------------------------
+// `X86Info` contains fields such as `vendor` and `brand_string` they are
+// represents characters in ASCII encoding. `vendor` length of characters is 13
```
gchatelet, line 38: "`X86Info` contains fields such as `vendor` and `brand_string` that are ASCII encoded strings."

src/impl_x86__base_implementation.inl
```diff
@@ -13,6 +13,65 @@
...
+// and `brand_string` is 49 (with null terminated string). We use
+// CPUID.1:E[D,C,B]X to get `vendor` and CPUID.8000_000[4:2]:E[D,C,B,A]X to get
+// `brand_string`
+//
+// Microarchitecture
+// -----------------------------------------------------------------------------
+// `GetX86Microarchitecture` function consists of check on vendor via
+// `IsVendorByX86Info`. We use `CPUID(family, model)` macro to define
+//  microarchitecture of vendor. In cases where the `family` and `model` is the
```
gchatelet, line 47: "to define the vendor's microarchitecture"

src/impl_x86__base_implementation.inl
```diff
@@ -13,6 +13,65 @@
...
+//  same for several microarchitectures we do a stepping check or in the worst
+//  case parsing `brand_string` as example `HasSecondFMA`. Details of
```
gchatelet, line 49: "or in the worst case we rely on parsing `brand_string` (see `HasSecondFMA` for an example)"

src/impl_x86__base_implementation.inl
```diff
@@ -13,6 +13,65 @@
...
+//  identification by `brand_string` can be found in the `Identification`
+//  section by reference:
+//  https://en.wikichip.org/wiki/intel/microarchitectures/cascade_lake
+//
+// CacheInfo X86
+// -----------------------------------------------------------------------------
+// We use common struct `CacheInfo` for all CPU architectures and this struct
```
gchatelet, line 56: "We use the `CacheInfo` struct to store information about cache levels. The maximum number of levels is hardcoded but can be increased if needed. We have full support .,."

src/impl_x86__base_implementation.inl
```diff
@@ -13,6 +13,65 @@
...
+// contains of `levels` with certain maximum number of cache levels, you can
+// feel free increase this value if more cache levels are needed. We have full
```
gchatelet, line 58: Maybe use bullet points to split between vendors, e.g.
 - Intel:
   - modern: 
   - old processors:
 - AMD:...

src/impl_x86__base_implementation.inl
```diff
@@ -13,6 +13,65 @@
...
+// we parse descriptors via `GetCacheLevelInfo`, see Application Note 485: Intel
+// Processor Identification and CPUID Instruction. For newer AMD and Intel
+// CPUs we use `ParseCacheInfo` function with difference of `leaf_id`.
+// Most other processors have the same approach as Intel and AMD, thus CPUs of
+// Hygon we detect via AMD cache detection and Zhaoxin via Intel.
+//
+// Internal structures
+// -----------------------------------------------------------------------------
+// We use internal structures such as `Leaves` and `OsPreserves` to hold cpuid
+// info and support of registers, since latency of `CPUID` instruction ~100
```
gchatelet, line 69: since latency of `CPUID` instruction **is around** ~100

src/impl_x86__base_implementation.inl
```diff
@@ -13,6 +13,65 @@
...
+// support of cache identification for Intel CPUs. For old processors of Intel
+// we parse descriptors via `GetCacheLevelInfo`, see Application Note 485: Intel
+// Processor Identification and CPUID Instruction. For newer AMD and Intel
+// CPUs we use `ParseCacheInfo` function with difference of `leaf_id`.
+// Most other processors have the same approach as Intel and AMD, thus CPUs of
+// Hygon we detect via AMD cache detection and Zhaoxin via Intel.
+//
+// Internal structures
+// -----------------------------------------------------------------------------
+// We use internal structures such as `Leaves` and `OsPreserves` to hold cpuid
```
gchatelet, line 68: "to hold cpuid" -> "to cache the result of `cpuid`"

gchatelet (commented):
I've added a bunch of documentation comments. Should be good to go then.

src/impl_x86__base_implementation.inl
```diff
@@ -13,6 +13,76 @@
...
+// • Hygon:
+//    we reuse AMD cache detection implementation.
+// • Zhaoxin:
+//    we reuse Intel cache detection implementation.
+//
+// Internal structures
+// -----------------------------------------------------------------------------
+// We use internal structures such as `Leaves` and `OsPreserves` to cache the
+// result of cpuid info and support of registers, since latency of CPUID
+// instruction is around ~100, see
```
gchatelet, line 80: ~100 **cycles**

gchatelet:
I'll squash and merge the PR once the typo is fixed. Thx a lot for the documentation!

gchatelet (approved):

Issue #200, opened 2021-11-14
list_cpu_features.exe does not detect SSE42 on Xeon X5650 (Windows)


RainbowMiner:
Here is the documentation of that specific CPU:
https://ark.intel.com/content/www/us/en/ark/products/47922/intel-xeon-processor-x5650-12m-cache-2-66-ghz-6-40-gts-intel-qpi.html

Here is the output of list_cpu_features.exe:
```
arch            : x86
brand           : Intel(R) Xeon(R) CPU           X5650  @ 2.67GHz
family          :   6 (0x06)
model           :  44 (0x2C)
stepping        :   2 (0x02)
uarch           : INTEL_WSM
flags           : aes,clfsh,cx16,cx8,dca,fpu,mmx,pclmulqdq,popcnt,smx,ss,sse,sse2,sse3,tsc
cache_info      : {"level":1,"cache_type":"data","cache_size":32768,"ways":8,"line_size":64,"tlb_entries":64,"partitioning":1},{"level":1,"cache_type":"instruction","cache_size":32768,"ways":4,"line_size":64,"tlb_entries":128,"partitioning":1},{"level":2,"cache_type":"unified","cache_size":262144,"ways":8,"line_size":64,"tlb_entries":512,"partitioning":1},{"level":3,"cache_type":"unified","cache_size":12582912,"ways":16,"line_size":64,"tlb_entries":12288,"partitioning":1}
```

If booted into Linux, sse4_2 is found. Partial output from cat /proc/cpuinfo:
```
processor       : 23
vendor_id       : GenuineIntel
cpu family      : 6
model           : 44
model name      : Intel(R) Xeon(R) CPU           X5650  @ 2.67GHz
stepping        : 2
microcode       : 0x1f
cpu MHz         : 2925.936
cache size      : 12288 KB
physical id     : 0
siblings        : 12
core id         : 10
cpu cores       : 6
apicid          : 21
initial apicid  : 21
fpu             : yes
fpu_exception   : yes
cpuid level     : 11
wp              : yes
[...]

PR #220, opened 2022-01-26
Fix list_cpu_features.exe does not detect SSE42 on Xeon X5650 (Windows)

toor1245:
https://github.com/google/cpu_features/issues/200


gchatelet:
I have a patch ready that embeds uarch detection inside `X86Info`. At first, I wanted to have it pushed before this one but I've already slowed you down enough. Let's submit this one as-is and I'll refactor it a bit afterwards.

gchatelet:
Fixes #200

Issue #221, opened 2022-01-28
Bazel build support


Vertexwahn:
Would be nice if there would also be build support for Bazel

keith:
looks like this was done in https://github.com/google/cpu_features/pull/222

PR #222, opened 2022-02-01
[NFC] Add bazel support to cpu_features

gchatelet:
Partial fix for #221

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,3 +1,5 @@
+workspace(name = "com_google_cpufeatures")
+
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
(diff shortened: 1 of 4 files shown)


gchatelet:
tested with `bazelisk-linux-amd64 test ...`

WORKSPACE
```diff
@@ -0,0 +1,21 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
Mizux, line 1: I would add at top:
```
workspace(name = "com_google_cpufeatures")
```

toor1245 (approved):

toor1245:
Do we need to provide "Build with Bazel" documentation or will it be a separate patch?

gchatelet:
@toor1245 good point ! Let's do it as a second step though, I suspect there will be a few iterations before everything works as correctly. We also need to setup continuous integration.

PR #226, opened 2022-02-20
Fix include paths in BUILD

pgavin:
Include paths should be specified through includes
rather than copts, so that they are handled correctly when
cpu_features is included in another project as an external repository.


gchatelet:
Thx for the PR @pgavin !

PR #232, opened 2022-03-09
ci: Add bazel jobs

Mizux:
* [optional] Use `git_repository()` to retrieve dependencies
* Add amd64 Linux Bazel based build jobs.

### DIFF
--- a/.github/workflows/amd64_bazel_linux.yml
+++ b/.github/workflows/amd64_bazel_linux.yml
@@ -1,0 +1,26 @@
+name: amd64 Bazel Linux
+
+on:
+  push:
+  pull_request:
+  schedule:
+    # min hours day(month) month day(week)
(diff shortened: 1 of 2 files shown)


Mizux:
MacOS and Windows are currently broken. I think it's an issue in our BUILD files not in jobs...

gchatelet:
As discussed offline let's:
 - remove the `build` part and keep only `test`
 - remove the `-c opt`
 - split the PR in three, one for Linux that we can submit right away and two branches for Mac and Windows so we can fix them

Mizux:
@gchatelet PTAL

note: I've just added the branch `bazel/windows` and `bazel/macos` feel free to commit in it to test/fix/iterate etc...

gchatelet:
LGTM Thx!

PR #244, opened 2022-06-21
Add macros for RISCV features

michael-roe:
Add macros that test which RISCV features have been enabled at compile time.

This patch has been tested with clang version 14.0.0 (and should also work with gcc).

clang does not currently support 128 bit integer registers (CPU_FEATURES_ARCH_RISCV128) or 128 bit floating point registers (CPU_FEATURES_COMPILED_RISCV_Q); all the other features are supported by clang 14.0.0.

These features are all either in the ratified version of the RISCV Unprivileged specification, or are ratified extensions. (the V vector extension, Zb bitmap extensions, and Zk crypto extensions were ratified in November 2021). i.e. they are all officially part of the RISCV spec.

### DIFF
--- a/include/cpu_features_macros.h
+++ b/include/cpu_features_macros.h
@@ -63,6 +63,22 @@
 #define CPU_FEATURES_ARCH_PPC
 #endif
 
+#if defined(__riscv)
+#define CPU_FEATURES_ARCH_RISCV
+#endif
+
+#if defined(__riscv) && defined(__riscv_xlen) && __riscv_xlen == 32
+#define CPU_FEATURES_ARCH_RISCV32
+#endif
+
+#if defined(__riscv) && defined(__riscv_xlen) && __riscv_xlen == 64
+#define CPU_FEATURES_ARCH_RISCV64
+#endif
+
+#if defined(__riscv) && defined(__riscv_xlen) && __riscv_xlen == 128
+#define CPU_FEATURES_ARCH_RISCV128
+#endif
+
 ////////////////////////////////////////////////////////////////////////////////
 // Os
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


include/cpu_features_macros.h
```diff
@@ -231,6 +247,84 @@
 #endif
 #endif  //  defined(CPU_FEATURES_ARCH_MIPS)
 
+#if defined(CPU_FEATURES_ARCH_RISCV)
+#if defined(__riscv_m)
+#define CPU_FEATURES_COMPILED_RISCV_M 1
```
toor1245, line 252: what about `__riscv_i`, `__riscv_`, `__riscv_p`, `__riscv_zfh`?
ref: https://github.com/riscv-non-isa/riscv-c-api-doc/blob/master/riscv-c-api.md

michael-roe: Thanks for the reply!

My reasoning got the omissions (which might not be good reasons) was
- I have some doubt as to whether RV32E embedded processors are capable enough to run the cpu_features library
- The p extension isn't ratified yet (and might never be)
- Zfh currently isn't supported by QEMU (although someone has submitted a patch), so although I can test compiler support, the next step (checking hwcap to see if the cpu you're running on supports it) is going to be difficult to test

i'll update this patch to add I, e, zfh and zfhmin. As p isn't ratified, I'd rather leave it out unless someone really wants it.

Mizux: Do you know which version of QEMU is needed ? so we can bump
https://github.com/google/cpu_features/blob/3c4801d12d9bec823d03a572f4f53cd42fa02b42/scripts/run_integration.sh#L36

michael-roe:
Ok, I've updated the PR to add macros for E, I, Zfh, Zfhmin, Zksed and Zksh.

clang 14.0.0 doesn't support rv32e, so you'll never see CPU_FEATURES_COMPILED_RISCV_E set unless you're using a different compiler.

toor1245:
looks good

toor1245 (approved):

PR #251, opened 2022-07-14
add windows ssse3,sse4_1,sse4_2 detection for non avx path

ajax16384:
This  PR allows to detect >sse3 for non-avx processors by using only windows methods. Which brings identical detection windows caps as of linux or macos.


ajax16384:
BTW this detection may superseed WESTMERE arch detect https://github.com/google/cpu_features/blob/8eb944f55d768a3c0250a1e80bb94de7b2560971/src/impl_x86_windows.c#L48
so it might be removed at all (and   _WIN32_WINNT compile time dependency seems fragile)

toor1245:
> BTW this detection may superseed WESTMERE arch detect
> 
> https://github.com/google/cpu_features/blob/8eb944f55d768a3c0250a1e80bb94de7b2560971/src/impl_x86_windows.c#L48
> 
> 
> so it might be removed at all (and _WIN32_WINNT compile time dependency seems fragile)

ah, my bad, when I was working on this problem I only referred to the Microsoft docs. You are right, there is no point in _WIN32_WINNT, it can be removed. Thanks for the PR 👍

test/cpuinfo_x86_test.cc
```diff
@@ -20,6 +20,20 @@
...
+#endif
+
+#if !defined(PF_SSE4_1_INSTRUCTIONS_AVAILABLE)
+#define PF_SSE4_1_INSTRUCTIONS_AVAILABLE            37
+#endif
+
+#if !defined(PF_SSE4_2_INSTRUCTIONS_AVAILABLE)
+#define PF_SSE4_2_INSTRUCTIONS_AVAILABLE            38
+#endif
+
```
toor1245, line 36: should we move this code to a common place to avoid duplication?

src/impl_x86_windows.c
```diff
@@ -43,6 +56,15 @@ static void DetectFeaturesFromOs(X86Info* info, X86Features* features) {
...
+      GetWindowsIsProcessorFeaturePresent(PF_SSSE3_INSTRUCTIONS_AVAILABLE);
+  features->sse4_1 =
+      GetWindowsIsProcessorFeaturePresent(PF_SSE4_1_INSTRUCTIONS_AVAILABLE);
+  features->sse4_2 =
+      GetWindowsIsProcessorFeaturePresent(PF_SSE4_2_INSTRUCTIONS_AVAILABLE);
+
+// do not bother checking PF_AVX*
+// cause AVX enabled processor will have XCR0 be exposed and this function will be skipped at all
 
 // https://github.com/google/cpu_features/issues/200
```
toor1245, line 69: this code can be removed as the new functionality replaces detection properly

test/cpuinfo_x86_test.cc
toor1245: no objections, @gchatelet what do you think?

ajax16384: ok

include/internal/windows_utils.h
```diff
@@ -0,0 +1,33 @@
+// Copyright 2017 Google LLC
```
toor1245, line 1: 2022?

test/cpuinfo_x86_test.cc
```diff
@@ -1173,15 +1176,13 @@ TEST_F(CpuidX86Test, WIN_INTEL_WESTMERE_EX) {
...
-  EXPECT_FALSE(info.features.sse4_2);
-#else
+  EXPECT_TRUE(info.features.sse);
+  EXPECT_TRUE(info.features.sse2);
+  EXPECT_TRUE(info.features.sse3);
   EXPECT_TRUE(info.features.ssse3);
   EXPECT_TRUE(info.features.sse4_1);
   EXPECT_TRUE(info.features.sse4_2);
-#endif
+
```
toor1245, line 1185: minor comment, extra space

ajax16384: ok

toor1245 (approved):

gchatelet:
Thx @ajax16384 and @toor1245 !

toor1245:
Created PR to update `IsProcessorFeaturePresent` docs https://github.com/MicrosoftDocs/sdk-api/pull/1251

PR #257, opened 2022-07-23
test enum macro consistency

ajax16384:
allows to check for LINE( presense in x86 feature and microarch arrays

### DIFF
--- a/test/cpuinfo_x86_test.cc
+++ b/test/cpuinfo_x86_test.cc
@@ -118,6 +118,30 @@ class CpuidX86Test : public ::testing::Test {
   }
 };
 
+TEST_F(CpuidX86Test, X86MicroarchitectureEnum) {
+   const char *last_name = GetX86MicroarchitectureName(X86_MICROARCHITECTURE_LAST_);
+   EXPECT_STREQ(last_name, "unknown microarchitecture");
+   for (int i = static_cast<int>(X86_UNKNOWN); i != static_cast<int>(X86_MICROARCHITECTURE_LAST_); ++i) {
+      const auto micro = static_cast<X86Microarchitecture>(i);
+      const char *name = GetX86MicroarchitectureName(micro);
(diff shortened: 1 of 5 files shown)


ajax16384:
might be extended to aarch64,arm,mips,ppc feature as well

toor1245:
> might be extended to aarch64,arm,mips,ppc feature as well

I think it is good validation and probably yes we should add this check for other archs.
@gchatelet, any objections?

test/cpuinfo_x86_test.cc
```diff
@@ -118,6 +118,30 @@ class CpuidX86Test : public ::testing::Test {
   }
 };
 
+TEST_F(CpuidX86Test, X86MicroarchitectureEnum) {
+   const char *last_name = GetX86MicroarchitectureName(X86_MICROARCHITECTURE_LAST_);
+   EXPECT_STREQ(last_name, "unknown microarchitecture");
+   for (int i = static_cast<int>(X86_UNKNOWN); i != static_cast<int>(X86_MICROARCHITECTURE_LAST_); ++i) {
+      const auto micro = static_cast<X86Microarchitecture>(i);
+      const char *name = GetX86MicroarchitectureName(micro);
+      EXPECT_FALSE(name == nullptr);
```
gchatelet, line 127: This one should be an `ASSERT_FALSE` (next line is UB if this one fails). Same in the function below.

gchatelet:
Yep that's a great idea. @ajax16384 can you add similar tests to other architectures as well?
Thx a lot for the PR.

test/cpuinfo_x86_test.cc
ajax16384: ok

ajax16384:
> Yep that's a great idea. @ajax16384 can you add similar tests to other architectures as well? Thx a lot for the PR.

added rest of arch features enum tests

PR #258, opened 2022-07-23
remove internal FillX86BrandString usage

ajax16384:
- get rid of deprecated funciton usage warning (though FillX86BrandString still available - let's not use it by ourself)
- add MSVC deprecation mark support


gchatelet:
Thx a lot @ajax16384 !

PR #256, opened 2022-07-22
add intel goldmont plus

ajax16384:

PR #259, opened 2022-07-26
Added some missing amd uarch

DaniAffCH:
According to https://en.wikichip.org/wiki/amd/cpuid I added detection of the following AMD uarch:

- ~Added family 0x18 model 0x00 to zen architecture~ 
- ~Moved family 0x17 model 0x18 from zen to zen plus architecture~
- ~Added family 0x15 model 0x10 to piledriver architecture~
- Created identifier for AMD_LLANO and added family 0x12 model 0x00 to it (Does it have to be merged with K12?)
   EDIT: merged with K12

### DIFF
--- a/src/impl_x86__base_implementation.inl
+++ b/src/impl_x86__base_implementation.inl
@@ -1801,7 +1801,7 @@ CacheInfo GetX86CacheInfo(void) {
   LINE(AMD_HAMMER)                  \
   LINE(AMD_K10)                     \
   LINE(AMD_K11)                     \
-  LINE(AMD_K12) /* K12 LLANO */     \
+  LINE(AMD_K12)                     \
   LINE(AMD_BOBCAT)                  \
   LINE(AMD_PILEDRIVER)              \
   LINE(AMD_STREAMROLLER)            \
--- a/src/impl_x86__base_implementation.inl
+++ b/src/impl_x86__base_implementation.inl
@@ -667,6 +667,9 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
       case CPUID(0x12, 0x01):
         // https://www.amd.com/system/files/TechDocs/44739_12h_Rev_Gd.pdf
         return AMD_K12;
+      case CPUID(0x12, 0x00):
+        // https://www.amd.com/system/files/TechDocs/44739_12h_Rev_Gd.pdf
+        return AMD_LLANO;
       case CPUID(0x14, 0x00):
       case CPUID(0x14, 0x01):
       case CPUID(0x14, 0x02):
@@ -676,6 +679,7 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
         // https://en.wikichip.org/wiki/amd/microarchitectures/bulldozer
         return AMD_BULLDOZER;
       case CPUID(0x15, 0x02):
+      case CPUID(0x15, 0x10):
       case CPUID(0x15, 0x11):
       case CPUID(0x15, 0x13):
         // https://en.wikichip.org/wiki/amd/microarchitectures/piledriver
@@ -695,11 +699,12 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
         return AMD_PUMA;
       case CPUID(0x17, 0x01):
       case CPUID(0x17, 0x11):
-      case CPUID(0x17, 0x18):
       case CPUID(0x17, 0x20):
+      case CPUID(0x18, 0x00):
         // https://en.wikichip.org/wiki/amd/microarchitectures/zen
         return AMD_ZEN;
       case CPUID(0x17, 0x08):
+      case CPUID(0x17, 0x18):
         // https://en.wikichip.org/wiki/amd/microarchitectures/zen%2B
         return AMD_ZEN_PLUS;
       case CPUID(0x17, 0x31):
@@ -1800,6 +1805,7 @@ CacheInfo GetX86CacheInfo(void) {
   LINE(AMD_K10)                     \
   LINE(AMD_K11)                     \
   LINE(AMD_K12)                     \
+  LINE(AMD_LLANO)                   \
   LINE(AMD_BOBCAT)                  \
   LINE(AMD_PILEDRIVER)              \
   LINE(AMD_STREAMROLLER)            \
--- a/src/impl_x86__base_implementation.inl
+++ b/src/impl_x86__base_implementation.inl
@@ -677,7 +677,6 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
         // https://en.wikichip.org/wiki/amd/microarchitectures/bulldozer
         return AMD_BULLDOZER;
       case CPUID(0x15, 0x02):
-      case CPUID(0x15, 0x10):
       case CPUID(0x15, 0x11):
       case CPUID(0x15, 0x13):
         // https://en.wikichip.org/wiki/amd/microarchitectures/piledriver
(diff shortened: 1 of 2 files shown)


src/impl_x86__base_implementation.inl
```diff
@@ -695,11 +699,12 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
         return AMD_PUMA;
       case CPUID(0x17, 0x01):
       case CPUID(0x17, 0x11):
-      case CPUID(0x17, 0x18):
       case CPUID(0x17, 0x20):
+      case CPUID(0x18, 0x00):
```
toor1245, line 703: This cpuid used only for Hygon processors and we already have in branch `IsVendorByX86Info(info, CPU_FEATURES_VENDOR_HYGON_GENUINE)`:
http://users.atw.hu/instlatx64/HygonGenuine/HygonGenuine0900F02_Hygon_CPUID.txt
http://users.atw.hu/instlatx64/HygonGenuine/HygonGenuine0900F11_Hygon_01_CPUID.txt
http://users.atw.hu/instlatx64/HygonGenuine/HygonGenuine0900F02_Hygon_CPUID3.txt
we define by vendor and this only applies to HygonGenuine

src/impl_x86__base_implementation.inl
```diff
@@ -667,6 +667,9 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
       case CPUID(0x12, 0x01):
         // https://www.amd.com/system/files/TechDocs/44739_12h_Rev_Gd.pdf
         return AMD_K12;
+      case CPUID(0x12, 0x00):
+        // https://www.amd.com/system/files/TechDocs/44739_12h_Rev_Gd.pdf
```
toor1245, line 671: I don't see LLANO by the link that you provided, so change it to https://en.wikichip.org/wiki/amd/cpuid, since there is a definition cpuid of LLANO

src/impl_x86__base_implementation.inl
```diff
@@ -667,6 +667,9 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
...
+        return AMD_LLANO;
```
toor1245, line 672: > Does it have to be merged with K12?
I'd leave it as `AMD_LLANO` since that's an accurate definition and we have a lot of cases at Intel so that's fine.

src/impl_x86__base_implementation.inl
```diff
@@ -695,11 +699,12 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
         return AMD_PUMA;
       case CPUID(0x17, 0x01):
       case CPUID(0x17, 0x11):
-      case CPUID(0x17, 0x18):
       case CPUID(0x17, 0x20):
+      case CPUID(0x18, 0x00):
         // https://en.wikichip.org/wiki/amd/microarchitectures/zen
         return AMD_ZEN;
       case CPUID(0x17, 0x08):
+      case CPUID(0x17, 0x18):
```
toor1245, line 707: you added detection for Picasso (ZEN+) but removed Banded Kestrel (ZEN), thus users that have Banded Kestrel will get incorrect information that is not supported by the processor

src/impl_x86__base_implementation.inl
```diff
@@ -667,6 +667,9 @@ X86Microarchitecture GetX86Microarchitecture(const X86Info* info) {
       case CPUID(0x12, 0x01):
         // https://www.amd.com/system/files/TechDocs/44739_12h_Rev_Gd.pdf
         return AMD_K12;
+      case CPUID(0x12, 0x00):
```
toor1245, line 670: could you add a test for this detection?

toor1245, line 670: could you add test?
http://users.atw.hu/instlatx64/AuthenticAMD/AuthenticAMD0300F10_K12_Llano_CPUID.txt

src/impl_x86__base_implementation.inl
toor1245, line 671: I checked other cpuid LLANO http://users.atw.hu/instlatx64/AuthenticAMD/AuthenticAMD0300F10_K12_Llano_CPUID2.txt:

```
CPUID 00000001: 00300F10-00020800-00802009-178BFBFF
```
Family 0x12, Model 0x01, thus you can move your cpuid to AMD_K12 or cpuid(0x12, 0x01) to AMD_LLANO

@gchatelet, what name should we leave?

gchatelet: We already have `K12` so we should reuse it to make sure we don't break someone.
So in the definition let's keep K12 with a comment `AMD_K12, // K12 LLANO` and here let's move the `CPUID` switch case up to return `AMD_K12`.

gchatelet:
Thank you @DaniAffCH for the PR (and @toor1245 for the review)

src/impl_x86__base_implementation.inl
DaniAffCH: I moved that cpuid according to wikichip, since they have the same cpuid `CPUID(0x17, 0x18)` it should be ZEN or ZEN+ not both. Do I leave ZEN as it is more general?

src/impl_x86__base_implementation.inl
DaniAffCH: I didn't find a suitable test for `0x00` in http://users.atw.hu/instlatx64/.

src/impl_x86__base_implementation.inl
toor1245: yup, you should leave ZEN

src/impl_x86__base_implementation.inl
toor1245: ok, just add a test for K12, I think it will be enough http://users.atw.hu/instlatx64/AuthenticAMD/AuthenticAMD0300F10_K12_Llano_CPUID.txt

DaniAffCH: It's already available marked as AMD_K12_LLANO

src/impl_x86__base_implementation.inl
```diff
@@ -1799,7 +1801,7 @@ CacheInfo GetX86CacheInfo(void) {
   LINE(AMD_HAMMER)                  \
   LINE(AMD_K10)                     \
   LINE(AMD_K11)                     \
-  LINE(AMD_K12)                     \
+  LINE(AMD_K12) /* K12 LLANO */     \
```
toor1245, line 1804: @gchatelet, did you mean to add a comment here or `include/cpuinfo_x86.h -> X86Microarchitecture`?

gchatelet: The latter `include/cpuinfo_x86.h`

src/impl_x86__base_implementation.inl
Mizux: +1 this link will be cool:
https://en.wikichip.org/wiki/amd/cpuid#Family_18_.2812h.29

src/impl_x86__base_implementation.inl
Mizux: resolved by: https://github.com/google/cpu_features/blob/f60b6f84051976e2cb226de7278c1d7edbb2b74e/test/cpuinfo_x86_test.cc#L351-L365

src/impl_x86__base_implementation.inl
DaniAffCH: No more needed as I removed LLANO identifier. It has been merged with K12

toor1245 (approved):

toor1245:
@DaniAffCH, I see the same case `CPUID(0x15, 0x10)` in https://github.com/google/cpu_features/pull/256, could you remove it?

PR #268, opened 2022-08-16
Move common logic of AArch64

toor1245:
To implement support aarch64 for Windows, macOS and FreeBSD I moved common logic to `src/impl_aarch64__base_implementation.inl` which is consistency to x86 base impl

### DIFF
--- a/src/impl_aarch64__base_implementation.inl
+++ b/src/impl_aarch64__base_implementation.inl
@@ -1,0 +1,84 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


Mizux (approved):

src/impl_aarch64__base_implementation.inl
```diff
@@ -0,0 +1,84 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cpuinfo_aarch64.h"
+
+#if !defined(CPU_FEATURES_ARCH_AARCH64)
+#error "Cannot compile cpuinfo_aarch64 on a non aarch64 platform.
+#endif
```
Mizux, line 19: your commit message should mention that you add this check too, if you can amend it... ;)

toor1245: thanks, updated :slightly_smiling_face:

Mizux (approved):
Thx !

PR #269, opened 2022-08-16
Get rid repeated branch

toor1245:
We have already `break` into else branch, thus we can remove the repeated branch

### DIFF
--- a/src/impl_x86__base_implementation.inl
+++ b/src/impl_x86__base_implementation.inl
@@ -1656,9 +1656,7 @@ static void ParseCacheInfo(const int max_cpuid_leaf, uint32_t leaf_id,
     const Leaf leaf = SafeCpuIdEx(max_cpuid_leaf, leaf_id, index);
     int cache_type_field = ExtractBitRange(leaf.eax, 4, 0);
     CacheType cache_type;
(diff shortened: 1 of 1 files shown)


src/impl_x86__base_implementation.inl
```diff
@@ -1656,9 +1656,7 @@ static void ParseCacheInfo(const int max_cpuid_leaf, uint32_t leaf_id,
     const Leaf leaf = SafeCpuIdEx(max_cpuid_leaf, leaf_id, index);
     int cache_type_field = ExtractBitRange(leaf.eax, 4, 0);
     CacheType cache_type;
-    if (cache_type_field == 0)
-      break;
-    else if (cache_type_field == 1)
+    if (cache_type_field == 1)
```
Mizux, line 1659: then line 1666 comment should be update since it would occur `if cache_type_field == 0` ;)

toor1245: What about:
```C
// Intel Processor Identification and the CPUID Instruction Application
// Note 485 page 37 Table 5-10. Deterministic Cache Parameters.
// We skip cache parsing in case null of cache type or cache type in the
// range of 4-31 according to documentation.
```

Issue #247, opened 2022-07-11
Add supoort for riscv64?


yuzibo:
Hi,
Do we have a plan to support riscv64?
I am willing to do it but maybe need your help.
I am stuck on how to define RiscvFeatures struct:
```
cat /proc/cpuinfo
processor       : 0
hart            : 2
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,bullet0

processor       : 1
hart            : 1
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,bullet0

processor       : 2
hart            : 3
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,bullet0

processor       : 3
hart            : 4
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,bullet0
```
I have a real riscv64 hardware by hand:)

toor1245:
Since riscv cpu identification instructions can only be performed in [M-mode](https://riscv.org/wp-content/uploads/2015/11/riscv-privived-spec-v1.7.pdf). Machine level is the highest level and as far as I understand we can't even run it in userland so the best approach is to parse proc/cpuinfo see Linux implementation https://github.com/google/cpu_features/blob/main/src/impl_x86_linux_or_android.c and the same approach we can use for FreeBSD https://github.com/google/cpu_features/blob/main/src/impl_x86_freebsd.c. if I'm not mistaken macOS and Windows don't support riscv

DaniAffCH:
@yuzibo Which board have you used? I would need this information to add a test

yuzibo:
Hi, @DaniAffCH
Thanks for working on that. My locally real riscv64 hardware is Unmatched boards.
```bash
Linux unmatched 5.18.0-2-riscv64 #1 SMP Debian 5.18.5-1 (2022-06-16) riscv64 GNU/Linux
```
So how can I help you to do test?
Or:
You can apply for  access to Unmatched boards' resources from here:
https://github.com/plctlab/riscv-lab-access
Any issue please let me know, thanks

DaniAffCH:
You could find the board name in the webpage where you bought it. It's just to give the test a meaningful name

yuzibo:
More info about Unmatched board is here:
https://www.sifive.com/boards/hifive-unmatched
The machines from plct is the same

DaniAffCH:
Thanks a lot! @yuzibo

toor1245:
[...]

PR #272, opened 2022-08-30
Add RISC-V support 

DaniAffCH:
I added RISC-V architecture support on linux by parsing `/proc/cpuinfo`. It has been tested using QEMU emulator.
I have done the following points: 
- features detection (i kept it as a single letter, is it fine or do i have to replace it with its description?)
- microarchitecture detection (i'm not sure if it's correct, so check it carefully)
- vendor detection
Tests have been added as well


src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+    if(!*info.vendor)
+      strcpy(info.vendor, "unknown_vendor");
+    if(!*info.uarch)
+      strcpy(info.uarch, "unknown_uarch");
+      
+    return info;
+}
+
+#endif  //  defined(CPU_FEATURES_OS_LINUX) || defined(CPU_FEATURES_OS_ANDROID)
+#endif  // CPU_FEATURES_ARCH_RISCV
```
toor1245, line 111: please add a new line at the end of the file and  remove `defined(CPU_FEATURES_OS_ANDROID)`

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+// Implementation.
+////////////////////////////////////////////////////////////////////////////////
+
+#include "internal/stack_line_reader.h"
+#include "internal/filesystem.h"
+
+#include <stdbool.h>
+#include <stdio.h>
```
toor1245, line 50: redundant include, please remove

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+// limitations under the License.
+
+#include "cpu_features_macros.h"
+
+#ifdef CPU_FEATURES_ARCH_RISCV
+#if defined(CPU_FEATURES_OS_LINUX)
+
+#include "cpuinfo_riscv.h"
+
+////////////////////////////////////////////////////////////////////////////////
```
toor1245, line 22: better to move duplicate logic to a generic place for other operating systems

include/internal/hwcaps.h
```diff
@@ -177,6 +177,9 @@ CPU_FEATURES_START_CPP_NAMESPACE
 #endif
 
 // https://elixir.bootlin.com/linux/latest/source/arch/riscv/include/uapi/asm/hwcap.h
+#define RISCV_HWCAP_32 0x32
+#define RISCV_HWCAP_64 0x64
+#define RISCV_HWCAP_128 0x128
```
toor1245, line 182: what is the purpose of these definitions and why represented in hex? Look like some magic numbers 
`0x128 -> 296` and I don't see that was declared by https://elixir.bootlin.com/linux/latest/source/arch/riscv/include/uapi/asm/hwcap.h

DaniAffCH: Actually they are a placeholder to use introspection table. The hex number are invented. How can i define introspection table's line without hwcap?

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
+// Copyright 2022 Google LLC
```
toor1245, line 1: please apply clang-format

include/internal/hwcaps.h
toor1245: you can make the same as X86 impl
```C
LINE(X86_FPU, fpu, , , ) 
```

include/cpuinfo_riscv.h
```diff
@@ -0,0 +1,72 @@
...
+    int m : 1;        // Integer Multiplication and Division
+    int v : 1;        // Vector Operations
+    int q : 1;        // Quad Precision Floating Point
+} RiscvFeatures;
+
+
+typedef struct {
+    RiscvFeatures features;
+    char          uarch[64];  // 0 terminated string
+    char          vendor[64]; // 0 terminated string
```
toor1245, line 46: I think we should consider API that will be consistent with other operating systems, what about FreeBSD
and possible future operating systems? Can we get this info?

include/cpuinfo_riscv.h
```diff
@@ -0,0 +1,72 @@
+// Copyright 2022 Google LLC
```
toor1245, line 1: please apply clang-format

test/cpuinfo_riscv_test.cc
```diff
@@ -0,0 +1,123 @@
...
+    EXPECT_FALSE(info.features.e);
+    EXPECT_TRUE(info.features.f);
+    EXPECT_TRUE(info.features.i);
+    EXPECT_TRUE(info.features.m);
+    EXPECT_FALSE(info.features.q);
+    EXPECT_FALSE(info.features.v);
+}
+
+}  // namespace
+}  // namespace cpu_features
```
toor1245, line 123: new line

src/impl_riscv_linux.c
DaniAffCH: At the moment there isn't a duplicate logic except for the `GetRiscvInfo()` function (just a few lines of code). Do I nevertheless have to move it to another file?

include/cpuinfo_riscv.h
DaniAffCH: I'm not sure about other os compatibility. The more general way would be using `misa` instruction but as we have seen in #247 it's not possible. Do you have any suggestion?

toor1245: I mean that we should define only common info, and avoid features which on certain operating systems will return empty results.

for example, if both Linux and FreeBSD return `hart` from the cpuinfo file, then we can add to our functionality, but if only one Linux/FreeBSD returns some info and the other doesn't provide that info we should avoid it.

src/impl_riscv_linux.c
toor1245: yes, but the next developer who will add support for another OS should rewrite/move your common logic

michael-roe (changes requested):
I think your parser of the RISC-V ISA string is incorrect, because the ISA string can contain multi-letter extensions, such as "_zbc". Simply searching for single letters within the ISA string won't work, you need to parse it to find out where the multi-letter extensions begin.

DaniAffCH:
Ok @michael-roe, Do I have to support [these extensions](https://wiki.riscv.org/display/HOME/Recently+Ratified+Extensions) as well?

michael-roe:
> Ok @michael-roe, Do I have to support [these extensions](https://wiki.riscv.org/display/HOME/Recently+Ratified+Extensions) as well?

Probably ought to support all the user space ones. The supervisor state ones (starting with S) are less relevant to user space  programs that call cpu_features, so can probably leave those out.

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+  StringView line = result.line;
+  StringView key, value;
+  if (CpuFeatures_StringView_GetAttributeKeyValue(line, &key, &value)) {
+    if (CpuFeatures_StringView_IsEquals(key, str("isa"))) {
+      StringView tmp_sv;
+      value.ptr+=2; // ignore 'rv' prefix at the beginning
+      value.size-=2;
+      for (size_t i = 0; i < RISCV_LAST_; ++i){
+        tmp_sv.ptr = kCpuInfoFlags[i];
+        tmp_sv.size = strlen(kCpuInfoFlags[i]);
```
toor1245, line 65: Why are you using libc functions like `strlen`, `strcpy`? According to documentation `cpu_features is suitable for implementing fundamental libc functions like malloc, memcpy, and memcmp`. Hence we use our implementation, see `src/copy.inl`, `src/equals.inl`.

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+
+static bool HandleRiscVLine(const LineResult result,
+                            RiscvInfo* const info) {
+  StringView line = result.line;
+  StringView key, value;
+  if (CpuFeatures_StringView_GetAttributeKeyValue(line, &key, &value)) {
+    if (CpuFeatures_StringView_IsEquals(key, str("isa"))) {
+      StringView tmp_sv;
+      value.ptr+=2; // ignore 'rv' prefix at the beginning
+      value.size-=2;
```
toor1245, line 62: please reuse exist functionality of StringView, which also includes checks

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+        tmp_sv.size = strlen(kCpuInfoFlags[i]);
+        kSetters[i](&info->features,
+                    CpuFeatures_StringView_IndexOf(value, tmp_sv) != -1);
+      
+      }
+    }
+    if (CpuFeatures_StringView_IsEquals(key, str("uarch"))){
+      int separatorIdx = CpuFeatures_StringView_IndexOfChar(value, ',');
+      StringView vendor = {.ptr = value.ptr, .size = separatorIdx};
+      StringView uarch = {.ptr = value.ptr + separatorIdx + 1, .size = value.size - separatorIdx - 1};
```
toor1245, line 74: same here, you should reuse StringView functionality

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+}
+
+RiscvInfo GetRiscvInfo(void){
+    RiscvInfo info = kEmptyRiscvInfo;
+    FillProcCpuInfoData(&info);
+
+    if(!*info.vendor)
+      strcpy(info.vendor, "unknown_vendor");
+    if(!*info.uarch)
+      strcpy(info.uarch, "unknown_uarch");
```
toor1245, line 105: @DaniAffCH, @michael-roe, Is it a real case that proc/cpuinfo will not display the `uarch` field?

DaniAffCH: Yes it is i guess. Since i couldn't find `/proc/cpuinfo` samples online, i asked on Reddit and i received [this one](https://www.reddit.com/r/RISCV/comments/wfz050/comment/iiwnmpd/?utm_source=share&utm_medium=web2x&context=3) that doesn't have the `uarch` field

DaniAffCH: If it is significant, qemu emulation doesn't provide that field as well

src/impl_riscv_linux.c
toor1245: @gchatelet, it seems I misunderstood. Is this the right point? Since I found several codes where `memmove` and `strlen` are used:
https://github.com/google/cpu_features/blob/main/src/stack_line_reader.c#L59 
https://github.com/google/cpu_features/blob/main/include/internal/string_view.h#L47

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+
+static const RiscvInfo kEmptyRiscvInfo;
+
+static bool HandleRiscVLine(const LineResult result,
+                            RiscvInfo* const info) {
+  StringView line = result.line;
+  StringView key, value;
+  if (CpuFeatures_StringView_GetAttributeKeyValue(line, &key, &value)) {
+    if (CpuFeatures_StringView_IsEquals(key, str("isa"))) {
+      StringView tmp_sv;
```
gchatelet, line 60: `if(CpuFeatures_StringView_StartsWith(value, "rv") == false) continue;`

src/impl_riscv_linux.c
gchatelet, line 62: value = CpuFeatures_StringView_PopFront(value, 2)

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+      value.ptr+=2; // ignore 'rv' prefix at the beginning
+      value.size-=2;
+      for (size_t i = 0; i < RISCV_LAST_; ++i){
+        tmp_sv.ptr = kCpuInfoFlags[i];
```
gchatelet, line 64: `StringView tmp_sv = str(kCpuInfoFlags[i]);`

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+        tmp_sv.size = strlen(kCpuInfoFlags[i]);
+        kSetters[i](&info->features,
+                    CpuFeatures_StringView_IndexOf(value, tmp_sv) != -1);
```
gchatelet, line 67: extract a variable:
```
bool is_set = CpuFeatures_StringView_IndexOf(value, tmp_sv) != -1;
kSetters[i](&info->features, is_set);
```

src/impl_riscv_linux.c
gchatelet: It used to be strongly enforced but it is not anymore as it is hard to prevent the compiler from generating calls to these functions. That is, even if the code contains no call to `C` memory functions, the compiler may decide to generate them (e.g., https://godbolt.org/z/h5KKM7MEj)

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+      
+      }
+    }
+    if (CpuFeatures_StringView_IsEquals(key, str("uarch"))){
+      int separatorIdx = CpuFeatures_StringView_IndexOfChar(value, ',');
+      StringView vendor = {.ptr = value.ptr, .size = separatorIdx};
```
gchatelet, line 73: What if `separatorIdx` is `-1`?

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+    
+    CpuFeatures_CloseFile(fd);
+  }
+}
+
+RiscvInfo GetRiscvInfo(void){
+    RiscvInfo info = kEmptyRiscvInfo;
+    FillProcCpuInfoData(&info);
+
+    if(!*info.vendor)
```
gchatelet, line 102: if(info.vendor[0] == '\0')
  CpuFeatures_StringView_CopyString(info.vendor, sizeof(info.vendor), str("unknown_vendor"));

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+      strcpy(info.vendor, "unknown_vendor");
+    if(!*info.uarch)
```
gchatelet, line 104: same here

src/impl_riscv_linux.c
gchatelet, line 111: Add an empty line at EOF

src/utils/list_cpu_features.c
```diff
@@ -408,6 +413,12 @@ static Node* CreateTree(void) {
   AddMapEntry(root, "microarchitecture",
               CreateString(strings.type.base_platform));
   AddFlags(root, &info.features);
+#elif defined(CPU_FEATURES_ARCH_RISCV)
+  const RiscvInfo info = GetRiscvInfo();
+  AddMapEntry(root, "arch", CreateString("risc-v"));
```
gchatelet, line 418: With enough time people will rely on this string. Are we sure we want "risc-v"?
Alternative naming scheme I can think of :
 - "riscv"
 - "riscv-32"/ "riscv-64" / "riscv-128"

test/cpuinfo_riscv_test.cc
```diff
@@ -0,0 +1,123 @@
...
+    EXPECT_TRUE(info.features.i);
+    EXPECT_TRUE(info.features.m);
+    EXPECT_FALSE(info.features.q);
+    EXPECT_FALSE(info.features.v);
+}
+
+TEST(CpuinfoRiscvTest, KendryteFromCpuInfo) {
+    ResetHwcaps();
+    auto& fs = GetEmptyFilesystem();
+    fs.CreateFile("/proc/cpuinfo", R"(hart  : 0
```
gchatelet, line 53: put `hart : 0` on a new line for visibility

test/cpuinfo_riscv_test.cc
gchatelet, line 123: new line at EOF

src/impl_riscv_linux.c
```diff
@@ -0,0 +1,111 @@
...
+static bool HandleRiscVLine(const LineResult result,
+                            RiscvInfo* const info) {
+  StringView line = result.line;
+  StringView key, value;
+  if (CpuFeatures_StringView_GetAttributeKeyValue(line, &key, &value)) {
+    if (CpuFeatures_StringView_IsEquals(key, str("isa"))) {
+      StringView tmp_sv;
+      value.ptr+=2; // ignore 'rv' prefix at the beginning
+      value.size-=2;
+      for (size_t i = 0; i < RISCV_LAST_; ++i){
```
gchatelet, line 63: The complexity here is O(N^2)
I'd rather scan the characters after "rv" and check if the character corresponds to one of the riscv extensions.

gchatelet (changes requested):

src/impl_riscv_linux.c
toor1245: should we provide info to REAMDE that even with compiler flags such as `freestanding` `no-stdlib` `no-builtin` there is a possibility that the compiler will generate C functions?

src/impl_riscv_linux.c
toor1245: `if(!CpuFeatures_StringView_StartsWith(value, "rv")) continue;`?

src/utils/list_cpu_features.c
toor1245: the second naming scheme seems better to me since this functionality is implemented with `GetRiscvInfo` and more precise arch detection compared to riscv, but I also don't mind `"riscv"`

include/cpuinfo_riscv.h
```diff
@@ -0,0 +1,72 @@
...
+typedef struct {
+    RiscvFeatures features;
+    char          uarch[64];  // 0 terminated string
+    char          vendor[64]; // 0 terminated string
+} RiscvInfo;
+
+typedef enum {
+    RISCV_32,
+    RISCV_64,
+    RISCV_128,
```
toor1245, line 52: is it right that we mixed features and arch to `RiscvFeaturesEnum` should we separate them?

README.md
```diff
@@ -172,14 +172,14 @@ flags           : aes,avx,cx16,smx,sse4_1,sse4_2,ssse3
...
-| Windows | yes² | not yet | not yet | N/A     | N/A     |
-| FreeBSD | yes² | not yet | not yet | not yet | not yet |
+|         | x86³ | ARM     | AArch64 | MIPS⁴   | POWER   | RISC-V  |
+|---------|:----:|:-------:|:-------:|:-------:|:-------:|:-------:|
+| Android | yes² | yes¹    | yes¹    | yes¹    | N/A     | N/A     |
+| iOS     | N/A  | not yet | not yet | N/A     | N/A     | N/A     |
+| Linux   | yes² | yes¹    | yes¹    | yes¹    | yes¹    | yes¹    |
+| MacOs   | yes² | N/A     | not yet | N/A     | no      | N/A     |
+| Windows | yes² | not yet | not yet | N/A     | N/A     | N/A     |
+| FreeBSD | yes² | not yet | not yet | not yet | not yet | N/A     |
```
toor1245, line 182: FreeBSD - `not yet`

src/impl_riscv_linux.c
gchatelet: Yep I think so, here is an example where the compiler generates a call even with all the compiler flags https://godbolt.org/z/rTTfzEq15

In fact, compilers [require some C functions to exist](https://gcc.gnu.org/onlinedocs/gcc/Standards.html#:~:text=GCC%20requires%20the%20freestanding%20environment%20provide%20memcpy%2C%20memmove%2C%20memset%20and%20memcmp.).

src/utils/list_cpu_features.c
gchatelet: As an afterthought, it seems that the 32 / 64 / 128 address space is really seen as a feature so we could stick to "risc-v". As you said the information exists elsewhere. I don't mind either, I just wanted to check whether we could regret this choice later on.

include/cpuinfo_riscv.h
```diff
@@ -0,0 +1,72 @@
...
+    int a : 1;        // Atomic Instructions
+    int c : 1;        // Compressed Instructions
+    int d : 1;        // Double Precision Floating Point
+    int e : 1;        // Base Integer Reduced
+    int f : 1;        // Single Precision Floating Point
+    int i : 1;        // Base Integer
+    int m : 1;        // Integer Multiplication and Division
+    int v : 1;        // Vector Operations
+    int q : 1;        // Quad Precision Floating Point
+} RiscvFeatures;
```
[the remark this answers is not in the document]
toor1245: I agree with you to stick naming according to riscv documentation, thus we eliminate the confusion with `riscv32`/`riscv64`/`riscv128`, also we can add a comment with the isa definition (base/ext):

```C
typedef struct {
    int rvmwo : 1;  // Base ISA. Weak Memory Ordering
    int rv32i: 1;   // Base ISA. Base Integer Instruction Set, 32-bit
    ...
    int zicsr: 1;   // Extension ISA. Control and Status Register (CSR)
    ...
} RiscvFeatures;
```

gchatelet:
Hi @DaniAffCH are you still interested in contributing to this branch?
I can take it over if you want.

gchatelet:
> Hi @gchatelet, yes I'm still interested in contributing to this project but I have a really busy period, so I'm unable to fix the PR issues, I'll do it as soon as I'm free though. If you want to keep the PR list clean feel free to close my PR and I'll open it again when I'm ready.

Ok no worries. I've started #287 based on this PR to retain your original work (I hope you don't mind!). I'm now working on fixing the comments we've made. Thx to @Mizux we also have a running CI for RISC-V 💪.

DaniAffCH:
> > Hi @gchatelet, yes I'm still interested in contributing to this project but I have a really busy period, so I'm unable to fix the PR issues, I'll do it as soon as I'm free though. If you want to keep the PR list clean feel free to close my PR and I'll open it again when I'm ready.
> 
> Ok no worries. I've started #287 based on this PR to retain your original work (I hope you don't mind!). I'm now working on fixing the comments we've made. Thx to @Mizux we also have a running CI for RISC-V muscle.

Ok, good job! I can probably be back active in 2 weeks. Please keep track of the problems you have fixed (or otherwise I'm gonna figure it out from your commits).

gchatelet:
Merged as part of #287 . Thx a lot @DaniAffCH for the initial implementation and research, this helped !

PR #274, opened 2022-09-19
S390X Support

marquitos0119:
This PR adds support for cpu_features on S390X architectures including the latest z16 hwcapabilities.


borntraeger:
z15 Model T01 would be 8561 and not  3932
you might also want some older processors as well

2964 and 2965 z13                
3906 and 3907 z14

marquitos0119:
Thanks for the review. I ended up removing the explicit model checking, since the elf [AT_PLATFORM](https://github.com/torvalds/linux/blob/a1901b464e7e3e28956ae7423db2847dbbfb5be8/arch/s390/kernel/processor.c#L256) definition includes this.

include/cpuinfo_s390x.h
```diff
@@ -0,0 +1,109 @@
...
+#ifndef CPU_FEATURES_INCLUDE_CPUINFO_S390X_H_
+#define CPU_FEATURES_INCLUDE_CPUINFO_S390X_H_
+
+#include "cpu_features_cache_info.h"
+#include "cpu_features_macros.h"
+
+CPU_FEATURES_START_CPP_NAMESPACE
+
+typedef struct {
+  int esan3: 1;
```
gchatelet, line 24: Can you add the documentation on the fields here instead of the enum constants below?

include/cpuinfo_s390x.h
```diff
@@ -0,0 +1,109 @@
...
+} S390XInfo;
+
+S390XInfo GetS390XInfo(void);
+
+typedef struct {
+  char platform[64];       // 0 terminated string
+} S390XPlatformTypeStrings;
+
+typedef struct {
+  char num_processors[64];       // 0 terminated string
```
gchatelet, line 62: What's the format of this string?

include/cpuinfo_s390x.h
```diff
@@ -0,0 +1,109 @@
...
+  S390XPlatformTypeStrings type;
+} S390XPlatformStrings;
+
+S390XPlatformStrings GetS390XPlatformStrings(void);
+
+////////////////////////////////////////////////////////////////////////////////
+// Introspection functions
+
+typedef enum {
+  S390_ESAN3,        /* instructions named N3, "backported" to esa-mode */
```
gchatelet, line 72: Please keep the comments as `//`

gchatelet (changes requested):
Thx a lot for the PR. A few minor comments before approving.

test/cpuinfo_s390x_test.cc
```diff
@@ -0,0 +1,86 @@
+// Copyright 2018 IBM.
```
toor1245, line 1: minor comment: could you make the copyright comment consistent? I see that in above files `Copyright 2022 IBM`

test/cpuinfo_s390x_test.cc
```diff
@@ -0,0 +1,86 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cpuinfo_s390x.h"
+
+#include "filesystem_for_testing.h"
+#include "gtest/gtest.h"
+#include "hwcaps_for_testing.h"
+#include "internal/string_view.h"
```
toor1245, line 20: I see that this include is unused, could you remove it?

include/cpuinfo_s390x.h
marquitos0119: This is just the line in /proc/cpuinfo that shows number of processors in form "# processors    : <processors>"

marquitos0119:
Thanks, I've made those changes

include/cpuinfo_s390x.h
gchatelet: Hmm it's probably best to get this field in a usable form.
I would make it an `int num_processors; // -1 if N/A`.

gchatelet:
> Thanks, I've made those changes

Thx, can you also turn the `num_processors` field into an `int`? then I think it's good to go.

include/cpuinfo_s390x.h
marquitos0119: Ah right, that makes sense

marquitos0119:
Made a rookie mistake, forgot to set my git user+email on a new environment and it failed the CLA check. I amended and I think its good to go now

test/cpuinfo_s390x_test.cc
```diff
@@ -0,0 +1,84 @@
...
+TEST(CpustringsS390XTest, z15) {
+  ResetHwcaps();
+  auto& fs = GetEmptyFilesystem();
+  fs.CreateFile("/proc/cpuinfo",
+                R"(vendor_id       : IBM/S390
+# processors    : 2
+bogomips per cpu: 24038.00
+max thread id   : 1
+features    : esan3 zarch stfle msa ldisp eimm dfp edat etf3eh highgprs te vx 
+vxd vxe gs vxe2 vxp sort dflt sie)");
```
gchatelet, line 76: Is the new line here expected?

gchatelet: Yes you must match `/proc/cpuinfo`, the parser expects the flags to be on a single line. For instance, if someone adds a test for `vxp` later on it won't be picked up in tests, although it would pick it up in production.

marquitos0119: Ok, that makes sense

PR #275, opened 2022-09-20
Download googletest only if testing is enabled

data-man:


CMakeLists.txt
```diff
@@ -163,8 +163,8 @@ endif()
 # tests
 #
 
-include(CTest)
 if(BUILD_TESTING)
+  include(CTest)
```
toor1245, line 167: @Mizux, currently it defaults to `ON` if the user doesn't specify `BUILD_TESTTING=ON/OFF`
https://github.com/Kitware/CMake/blob/master/Modules/CTest.cmake#L50
Is this by design or should we apply these changes?

Mizux: yup is by design, here we would like to follow the idiomatic CMake way i.e. use `include(CTest)` which provide/advertise the option `BUILD_TESTING` **and** also call `enable_testing()`.
So this call must be placed **before** the if statement.

Concerning the default to `ON`, our point is we follow the CTest default behaviour so if CTest set it ON it will be ON and can be changed if the CTest implementation change one day... again no surprise here for a distro maintainer accustomed to the CMake build system default behaviour...

ref: https://cmake.org/cmake/help/latest/module/CTest.html#module:CTest
ref: https://cmake.org/cmake/help/book/mastering-cmake/cmake/Help/guide/tutorial/index.html#adding-support-for-a-dashboard-step-8

data-man:
@gchatelet
Why did you close this PR? Didn't you understand the problem?

gchatelet:
@data-man I didn't mean to be rude but as @Mizux [described](https://github.com/google/cpu_features/pull/275#discussion_r1002999123) this is work as intended.
@Mizux is there something we can do to prevent downloading googletest when tests are disabled?

Mizux:
`include(CTest)` expose the option `BUILD_TESTING` (ON by default) if you want to disable test simply pass `-DBUILD_TESTING_ON` **as usual**.

Concerning  googletest it is guarded inside the if so it won't be require if test are disabled.
One thing we could rewrite is using FetchContent to pull googletest but not sure is possible with the minimum required cmake version...

ref
https://github.com/google/cpu_features/blob/26852665b4f1bdd14f10deb6d3201018cebaf2b2/CMakeLists.txt#L176-L210

PR #282, opened 2022-10-25
Add REP instructions detection

toor1245:
In this PR I added the detection of rep instructions:
* Fast Short REP MOV
* Fast zero-length REP MOVSB
* Fast short REP STOSB
* Fast short REP CMPSB/SCASB

image

_Note:_
I didn't find cpuid dump with REP MOVSB and REP CMPSB/SCASB only added alder lake where we have Fast short REP STOSB/ Fast Short REP MOV


Mizux (approved):
LGTM

PR #284, opened 2022-11-02
Update README.md to report support of POWER and s390x

gchatelet:


Mizux (approved):
LGTM

README.md
```diff
@@ -172,14 +176,14 @@ flags           : aes,avx,cx16,smx,sse4_1,sse4_2,ssse3
...
-| Linux   | yes² | yes¹    | yes¹    | yes¹    | yes¹    |
-| MacOs   | yes² | N/A     | not yet | N/A     | no      |
-| Windows | yes² | not yet | not yet | N/A     | N/A     |
-| FreeBSD | yes² | not yet | not yet | not yet | not yet |
+|         | x86³ | AArch64 | ARM     | MIPS⁴   | s390x   | POWER   |
+|---------|:----:|:-------:|:-------:|:-------:|:-------:|:-------:|
+| Linux   | yes² | yes¹    | yes¹    | yes¹    | yes¹    | yes¹    |
+| FreeBSD | yes² | not yet | not yet | not yet | not yet | not yet |
+| MacOs   | yes² | not yet | N/A     | N/A     | no      | no      |
+| Windows | yes² | not yet | not yet | N/A     | N/A     | N/A     |
```
toor1245, line 184: there is a question related to Windows AArch64, we have two ways to add support via software driver as we need EL1 (userland) to get info about system registers, see https://github.com/google/cpu_features/pull/186 or just read via `IsProcessorFeaturePresent` and `HARDWARE\\DESCRIPTION\\System\\CentralProcessor\\0`

the first approach is not easy, but we will have consistent features detection with Linux/FreeBSD, 
the second approach is easy to get info, but `IsProcessorFeaturePresent` provides a small range of features detection 

By the way, Linux and FreeBSD support reading features EL1 in user space (https://www.kernel.org/doc/html/latest/arm64/cpu-feature-registers.html), so we can replace impl to detection via asm if we want

Thoughts?

toor1245: also, there is an interesting fact in Windows arm64 that we can run our x86 impl and get virtual info :) 

[Broadcom BCM2837B0 (Raspberry Pi 3B+)](http://users.atw.hu/instlatx64/Virtual_CPU_/Virtual%20CPU%200000F4A_BCM2837B0_CPUID.txt)

gchatelet: For the purpose of this PR I will leave it as-is but we definitely want to support other ways to gather CPU info data. For instance, getting `midr` is a nice addition, it's a bit like the `vendor` and `brand_string` for x86.

That said, I really want to keep `cpu_features` as a userland library ( i.e., not a driver ). EL1 reading via filesystem under Linux / FreeBSD LGTM. Writing a driver for Windows less so.

(The raspberry pi emulating x86 on windows is amazing 😀)

gchatelet: I'll close this PR but we can continue the discussion here if you want.

toor1245: Got it, so we choose reading from `IsProcessorFeaturePresent`, my next question is about public API should we create a separate struct for win arm64 so as not to confuse the user (for instance `Aarch64WindowsInfo`)? Since for the current struct `Aarch64Info` we have flags that `IsProcessorFeaturePresent` doesn't provide, for example, `PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE` we don't know which instructions CPU supports AES, SHA1, SHA2...

toor1245: I'll leave here FreeBSD RPI4 `dmesg.boot`:
```ini
CPU  0: ARM Cortex-A72 r0p3 affinity:  0
                   Cache Type = <64 byte D-cacheline,64 byte I-cacheline,PIPT ICache,64 byte ERG,64 byte CWG>
 Instruction Set Attributes 0 = <CRC32>
 Instruction Set Attributes 1 = <>
 Instruction Set Attributes 2 = <>
         Processor Features 0 = <AdvSIMD,FP,EL3 32,EL2 32,EL1 32,EL0 32>
         Processor Features 1 = <>
      Memory Model Features 0 = <TGran4,TGran64,SNSMem,BigEnd,16bit ASID,16TB PA>
      Memory Model Features 1 = <8bit VMID>
      Memory Model Features 2 = <32bit CCIDX,48bit VA>
             Debug Features 0 = <DoubleLock,2 CTX BKPTs,4 Watchpoints,6 Breakpoints,PMUv3,Debugv8>
             Debug Features 1 = <>
         Auxiliary Features 0 = <>
         Auxiliary Features 1 = <>
AArch32 Instruction Set Attributes 5 = <CRC32,SEVL>
AArch32 Media and VFP Features 0 = <FPRound,FPSqrt,FPDivide,DP VFPv3+v4,SP VFPv3+v4,AdvSIMD>
AArch32 Media and VFP Features 1 = <SIMDFMAC,FPHP DP Conv,SIMDHP SP Conv,SIMDSP,SIMDInt,SIMDLS,FPDNaN,FPFtZ>
CPU  1: ARM Cortex-A72 r0p3 affinity:  1
CPU  2: ARM Cortex-A72 r0p3 affinity:  2
CPU  3: ARM Cortex-A72 r0p3 affinity:  3
```

toor1245: macOS hw docs: 
https://developer.apple.com/documentation/kernel/1387446-sysctlbyname/determining_instruction_set_characteristics

gchatelet: I forgot to hit send 😅
"If windows fields are just a subset of the fields I would just keep the same structure and document which fields are not populated on Windows. If they are disjoint (some exist on windows only, others exist on linux only) then it may make sense to have different structs. For now I would just keep the same and use documentation. We can always add a different struct later on if it makes sense. Keeping the interface as small as possible has maintenance value."

toor1245: clear, thanks a lot! I will implement it for Windows and will play with parsing FreeBSD if no objections

PR #285, opened 2022-11-02
Improve CI (Aarch64, Power)

Mizux:
Few updates

* cleanup scripts
* Add Aarch64 bootlin support (script, cmake/ci, workflows)
* Add Power bootlin support (script, cmake/ci, workflows)

### DIFF
--- a/scripts/run_integration.sh
+++ b/scripts/run_integration.sh
@@ -33,7 +33,7 @@ function install_qemu() {
     >&2 echo 'QEMU is disabled !'
     return 0
   fi
-  local -r QEMU_VERSION=${QEMU_VERSION:=5.2.0}
+  local -r QEMU_VERSION=${QEMU_VERSION:=7.1.0}
   local -r QEMU_TARGET=${QEMU_ARCH}-linux-user
 
   if echo "${QEMU_VERSION} ${QEMU_TARGET}" | cmp --silent "${QEMU_INSTALL}/.build" -; then
@@ -70,11 +70,7 @@ function install_qemu() {
     --disable-opengl \
     --disable-sdl \
     --disable-virglrenderer \
-    --disable-vte \
-    --enable-modules
-
-  # --static Not supported on Archlinux
-  # so we use --enable-modules
+    --disable-vte
 
   # wrapper on ninja
   make -j8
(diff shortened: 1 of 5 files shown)


scripts/run_integration.sh
```diff
@@ -145,6 +141,27 @@ function expand_bootlin_config() {
...
+    "ppc64le")
+      local -r POWER_URL="https://toolchains.bootlin.com/downloads/releases/toolchains/powerpc64le-power8/tarballs/powerpc64le-power8--glibc--stable-2021.11-1.tar.bz2"
+      local -r GCC_PREFIX="powerpc64le"
+      ;;
+    "ppc64")
+      local -r POWER_URL="https://toolchains.bootlin.com/downloads/releases/toolchains/powerpc64-power8/tarballs/powerpc64-power8--glibc--stable-2021.11-1.tar.bz2"
+      local -r GCC_PREFIX="powerpc64"
+      ;;
+    "ppc")
+      #local -r POWER_URL="https://toolchains.bootlin.com/downloads/releases/toolchains/powerpc-e500mc/tarballs/powerpc-e500mc--glibc--stable-2021.11-1.tar.bz2"
```
gchatelet, line 161: Why is it commented out?

scripts/run_integration.sh
```diff
@@ -230,12 +253,14 @@ function expand_codescape_config() {
     "mips64")
       MIPS_FLAGS="-EB -mips64r6 -mabi=64"
       FLAVOUR="mips-r6-hard"
+      #MIPS_FLAGS="-EB -mips64r2 -mabi=64"
```
gchatelet, line 256: same here and below, why do you need to keep these comments?

scripts/run_integration.sh
Mizux: While QEmu provide a single `ppc` target (as well as docker), bootlin provide several flavors
`440fp` and `e500mc` for PowerPC while here I arbitrarily the `440fp` we could have choose the other one...

An other way would be to have `ppc-e500mc` and `ppc-440fp` (and `ppc` alias of `ppc-440fp` ?)

ref:
```sh
docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
...
Setting /usr/bin/qemu-ppc-static as binfmt interpreter for ppc
Setting /usr/bin/qemu-ppc64-static as binfmt interpreter for ppc64
Setting /usr/bin/qemu-ppc64le-static as binfmt interpreter for ppc64le
```

note: same issue with `ppc64[le]`: `power8`, `e5500`, 'e6500', `le-power8`...

scripts/run_integration.sh
Mizux: same here maybe we could have a flag to decide which version to use so user can easily reuse this script to target the correct CPU by just changing one line, so while we only enable one flavor it's easy for the community to adapt.

Or again we should/could have both `mips-r6`/`mips-r2` and a `mips` alias to target the last one aka `mips-r6` ?

PR #286, opened 2022-11-02
Add RISC-V CI support

Mizux:
This follow(Fix) or precede(TDD):
* #272

This PR add:
* CI: Add riscv support
  * [x] Add bootlin toolchain in `scripts/run_integration.sh`
  * [x] Add riscv support in `cmake/ci/Makefile`
  * [x] Add `riskv_linux_cmake.yml` workflows
  * [x] Add RISCV to badge script (todo(gchatelet): rerun it to regenerate README.md)
* Fixup: variable fixup in expand_bootlin (ed `s/POWER_/TOOLCHAIN_/`)

note: I've sorted arch badges alphabetically...

PR #287, opened 2022-11-09
Support risc-v

gchatelet:
This is a retake of:
* #272,
* #286
and fixing 
* #247.


gchatelet:
I still need to address some of the original comments from #272.

gchatelet:
Let´s wait for #288 to be in before submitting this one so we can get the test ran through QEMU and make sure they pass.

Mizux (approved):

PR #290, opened 2023-01-21
cmake: make default test building dependent on being a main project build vs. being included into other projects

damageboy:
Hopefully this is self explanatory.
When using FetchContent with CMake/CPM this will prevent automatically discovering and building tests.
The option to force this is still preserved by proviing `BUILD_TESTING` as an externally settable option

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,14 @@ project(CpuFeatures VERSION 0.7.0 LANGUAGES C)
 
 set(CMAKE_C_STANDARD 99)
 
+# when cpu_features is included as subproject (i.e. using add_subdirectory(cpu_features))
+# in the source tree of a project that uses it, test rules are disabled.
+if(NOT CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+  option(BUILD_TESTING "Enable test rule" OFF)
+else()
+  option(BUILD_TESTING "Enable test rule" ON)
+endif()
+
 # Default Build Type to be Release
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING


CMakeLists.txt
```diff
@@ -10,6 +10,17 @@ project(CpuFeatures VERSION 0.7.0 LANGUAGES C)
 
 set(CMAKE_C_STANDARD 99)
 
+# Determine if cpu_features is built as a subproject (using add_subdirectory)
+# or if it is the master project.
+if (NOT DEFINED CPU_FEATURES_MAIN_PROJECT)
```
toor1245, line 15: @Mizux, should we add explanation of `CPU_FEATURES_MAIN_PROJECT` parameter to README?

Mizux: I would say no because it is just an internal temp variable result of the test `if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)`, so there is not intention to expose it to the user.

Concerning implementation I prefer the abseil-cpp one
https://github.com/abseil/abseil-cpp/blob/7e8d8018f621e94182876535320718542a4c5f09/CMakeLists.txt#L60-L66
while the option is duplicated it doesn't introduce a new `set()` variable...

```cmake
# when cpu_features is included as subproject (i.e. using add_subdirectory(cpu_features))
# in the source tree of a project that uses it, test rules are disabled.
if(NOT CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
  option(BUILD_TESTING "Enable test rule" OFF)
else()
  option(BUILD_TESTING "Enable test rule" ON)
endif()
```

gchatelet: @damageboy do you mind using @Mizux suggestion?

damageboy: Sorry, I totally missed this, I will fix according to @Mizux comment

Mizux (approved):

damageboy:
Can this be merged?

gchatelet:
Thx @damageboy for the PR (and for the ping!)

PR #293, opened 2023-01-28
Update C++ standard for tests from C++11 to C++14

musicinmybrain:
Since 1.13.0, gtest requires at least C++14. When it is automatically downloaded and compiled, gtest’s CMake scripts override the C++ standard, so the tests are compiled in C++14 mode already. Changing the C++ standard in cpu_features will make it easier for users such as Linux distribution packagers who are linking against an external or system-wide copy of gtest 1.13.0 or later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,9 +179,9 @@ if(BUILD_TESTING)
   # found.
   enable_language(CXX)
 
-  set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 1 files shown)


gchatelet:
I don't mind bumping the standard to c++14. @Mizux any opinion here?

Mizux:
Same here, most toolchain/distro supporting C++11 should also support C++14 now..

PR #303, opened 2023-03-20
Add some cmake options useful for embedding

amitin:
While following the [recommended](https://github.com/google/cpu_features/blob/main/cmake/README.md) way to incorporate cpu_features into a cmake project some cpu_features lib build artefacts become included into the target project. 
For example, list_cpu_features executable will be built and installed among project target binaries.

This pull request proposes to include some cmake options making cpu_features library embedding more flexible:
1. BUILD_SAMPLE_APP
This option controls if list_cpu_features will be built and included into the install target. Default is ON.
2. INCLUDE_INSTALL
This option allows to omit providing install targets, disabling installing cpu_features lib build artefacts into the main project install targets. Default in ON.
3. INCLUDE_TESTS
This option disables testing target for cpu_features at all. Since cpu_features uses standard cmake CTest BUILD_TESTING variable then it could interfere with the main project cmake settings. Having additional option solves this issue. Default is ON.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -223,24 +223,25 @@ cc_library(
         PLATFORM_CPU_RISCV32: ["src/impl_riscv_linux.c"],
         PLATFORM_CPU_RISCV64: ["src/impl_riscv_linux.c"],
     }),
-    copts = C99_FLAGS,
-    includes = INCLUDES,
-    textual_hdrs = selects.with_or({
+    hdrs = selects.with_or({
         PLATFORM_CPU_X86_64: [
-            "src/impl_x86__base_implementation.inl",
             "include/cpuinfo_x86.h",
             "include/internal/cpuid_x86.h",
             "include/internal/windows_utils.h",
         ],
         PLATFORM_CPU_ARM: ["include/cpuinfo_arm.h"],
-        PLATFORM_CPU_ARM64: [
-            "src/impl_aarch64__base_implementation.inl",
-            "include/cpuinfo_aarch64.h",
-        ],
+        PLATFORM_CPU_ARM64: ["include/cpuinfo_aarch64.h"],
         PLATFORM_CPU_MIPS: ["include/cpuinfo_mips.h"],
         PLATFORM_CPU_PPC: ["include/cpuinfo_ppc.h"],
         PLATFORM_CPU_RISCV32: ["include/cpuinfo_riscv.h"],
         PLATFORM_CPU_RISCV64: ["include/cpuinfo_riscv.h"],
+    }),
+    copts = C99_FLAGS,
+    includes = INCLUDES,
+    textual_hdrs = selects.with_or({
+        PLATFORM_CPU_X86_64: ["src/impl_x86__base_implementation.inl"],
+        PLATFORM_CPU_ARM64: ["src/impl_aarch64__base_implementation.inl"],
+        "//conditions:default": [],
     }) + [
         "src/define_introspection.inl",
         "src/define_introspection_and_hwcaps.inl",
@@ -285,10 +286,7 @@ cc_library(
             "include/internal/windows_utils.h",
         ],
         PLATFORM_CPU_ARM: ["include/cpuinfo_arm.h"],
-        PLATFORM_CPU_ARM64: [
-            "src/impl_aarch64__base_implementation.inl",
-            "include/cpuinfo_aarch64.h",
-        ],
+        PLATFORM_CPU_ARM64: ["include/cpuinfo_aarch64.h"],
         PLATFORM_CPU_MIPS: ["include/cpuinfo_mips.h"],
         PLATFORM_CPU_PPC: ["include/cpuinfo_ppc.h"],
         PLATFORM_CPU_RISCV32: ["include/cpuinfo_riscv.h"],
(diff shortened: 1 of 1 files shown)


toor1245:
I'm not sure about INCLUDE_TESTS, user can use `ExternalProject_Add` and pass `BUILD_TESTSING` with another paramater in `CMAKE_ARGS` like this: 
```cmake
include(ExternalProject)
ExternalProject_Add(cpu_features
  GIT_REPOSITORY    https://github.com/google/cpu_features.git
  GIT_TAG           main
  SOURCE_DIR        "${CMAKE_BINARY_DIR}/cpu_features-src"
  BINARY_DIR        "${CMAKE_BINARY_DIR}/cpu_features-build"
  CMAKE_ARGS
    -DBUILD_TESTING=OFF
)
```

@Mizux, what do you think?

amitin:
Well, I did try using 'external project' cmake module before, but actually this pulls out more problems, for example, it's not possible to use 'find_package' until cpu_features install target's done.
So I followed the recommended way of embedding cpu_features.

Mizux:
My 2 cents:

1. `BUILD_SAMPLE_APP`: ok on the principle to be able to disable this executable

2. `INCLUDE_INSTALL`: not sure is useful, which scenario need it ? 
maybe want to static link with libcpu_feature and integrate the sources using fetch or add_subdirectory ?

3. `INCLUDE_TESTS`: not OK, we already have BUILD_TESTING you can set to disable tests also in another PR test will be disable by default if cpu_feature is not the cmake root project

## side note

Now, `Fetch()` nicely interact with `find_package()` thanks to the 3.24 `OVERRIDE_FIND_PACKAGE`
see: https://cmake.org/cmake/help/latest/module/FetchContent.html#commands

Since variable are directory scoped you can add the `FetchContent()` in a subdir CMakeLists.txt and just before you add
`set(BUILD_TESTING OFF)`  thus all FetchContent will be run with `BUILD_TESTING=OFF` (even if it set to ON for the super project)

CMakeLists.txt
```diff
@@ -174,88 +184,95 @@ endif()
...
-  "${PROJECT_BINARY_DIR}/CpuFeaturesConfigVersion.cmake"
-  COMPATIBILITY SameMajorVersion
-)
-install(
-  FILES
+if(INCLUDE_INSTALL)
+  include(GNUInstallDirs)
+  set(install_targets cpu_features)
+  if(BUILD_SAMPLE_APP)
+    set(install_targets ${install_targets} list_cpu_features)
```
Mizux, line 245: why not using list(APPEND ...) ?

amitin: This is the same and I found it more readable, but I'll fix it in order to follow your code style.

amitin:
> 2\. `INCLUDE_INSTALL`: not sure is useful, which scenario need it ?
>     maybe want to static link with libcpu_feature and integrate the sources using fetch or add_subdirectory ?

Attached is the [sample app](https://github.com/google/cpu_features/files/11039926/cpu_feature_app.zip) which exactly reflects what we have in our project. This illustrates the problems I'm trying to solve with this pull request.

> 3\. `INCLUDE_TESTS`: not OK, we already have BUILD_TESTING you can set to disable tests also in another PR test will 

Excellent! I'll remove this from my PR and will wait for that PR to land.

> Now, `Fetch()` nicely interact with `find_package()` thanks to the 3.24 `OVERRIDE_FIND_PACKAGE`

Unfortunately, we can't currently use cmake 3.24, the latest version available to use is 3.13.

gchatelet:
Part of this PR has been submitted in https://github.com/google/cpu_features/pull/317. amitin@ do you still think this PR is necessary?

amitin:
> Part of this PR has been submitted in #317. @amitin do you still think this PR is necessary?

Yes. That PR only partially solves the problem of embedding which we have to get solved.
My PR also includes an option to disable any 'install' targets, so we don't need to install cpu_features include files and static lib.
Additionally, it's not really clear why you didn't land my PR which I submitted six month ago but landed a new one.

gchatelet:
> > Part of this PR has been submitted in #317. @amitin do you still think this PR is necessary?
> 
> Yes. That PR only partially solves the problem of embedding which we have to get solved. My PR also includes an option to disable any 'install' targets, so we don't need to install cpu_features include files and static lib.

AFAIU:
1. is solved by #317 
2. remains to be solved
3. is solved as per https://github.com/google/cpu_features/pull/303#issuecomment-1479495172

I'll have a look at 2. tomorrow.

> Additionally, it's not really clear why you didn't land my PR which I submitted six month ago but landed a new one.

My apologies for this, this PR went out of my radar and I started reviewing the most recent PRs first. It's unfortunate that #317 went in before yours.

Mizux:
For `2` and to be loosely consistent (and also since i like it) you could (should ?) follow the abseil-cpp style
https://github.com/abseil/abseil-cpp/blob/197bac8574c511b1937b60ce6405462f9af678ce/CMakeLists.txt#L65-L71
(note: prefer to use `ENABLE_INSTALL` since cpu_feature don't prefix its options)

note: https://cmake.org/cmake/help/latest/module/ExternalProject.html#explicit-step-management
> Install Step Options:
Passing an empty string as the <cmd> makes the install step do nothing.

Other proposal
https://github.com/google/googletest/blob/main/CMakeLists.txt#L20 INSTALL_GTEST (suffix)
https://github.com/protocolbuffers/protobuf/blob/main/CMakeLists.txt#L27 protobuf_INSTALL (prefix)
https://github.com/pybind/pybind11/blob/master/CMakeLists.txt#L86-L91 PYBIND11_INSTALL (prefix with master check)

amitin:
Opened https://github.com/google/cpu_features/pull/325 as I created another patch which includes the most recent changes related to my previous PR.

PR #304, opened 2023-03-29
Use system googletest dependency when available

topazus:


gchatelet:
@Mizux does that look good to you?

CMakeLists.txt
```diff
@@ -184,7 +184,8 @@ if(BUILD_TESTING)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   set(CMAKE_CXX_EXTENSIONS OFF) # prefer use of -std14 instead of -gnustd14
 
-  if(NOT TARGET gtest OR NOT TARGET gmock_main)
+  find_package(GTest REQUIRED)
+  if(NOT TARGET GTest::gtest OR NOT TARGET GTest::gmock_main)
```
Mizux, line 188: my 2 cents,

1. GTest::gtest seems to be available only since CMake 3.5 (cpu_feature still use 3.0 as minimum requirement...)
ref: https://cmake.org/cmake/help/v3.5/module/FindGTest.html
ref: https://cmake.org/cmake/help/v3.4/module/FindGTest.html

2. Googletest didn't provide these aliases until recently
issue: https://github.com/google/googletest/issues/2429
PR: https://github.com/google/googletest/pull/3155
commit: https://github.com/google/googletest/commit/6c5c4554ac218a8e19168edc121b1ad232015185
available in v1.14.0 … release-1.11.0
 
 3. If your superbuild is using `add_subdirectory()` OR `fetch_content()` (without the new flag `FIND_PACKAGE_ARGS NAMES ...`) then you `find_package()` won't work as expected (i.e. ignore the call since your super already provide it)
 ref: https://cmake.org/cmake/help/latest/module/FetchContent.html#integrating-with-find-package
 
 4. Also in 3.24 we would be able to have FetchContent calling find_package() before sourcing if unavailable...
 ref: https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_TRY_FIND_PACKAGE_MODE

CMakeLists.txt
```diff
@@ -184,7 +184,8 @@ if(BUILD_TESTING)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   set(CMAKE_CXX_EXTENSIONS OFF) # prefer use of -std14 instead of -gnustd14
 
-  if(NOT TARGET gtest OR NOT TARGET gmock_main)
+  find_package(GTest REQUIRED)
```
Mizux, line 187: since you add `REQUIRED` -> if not found the cmake configure will fail where instead you would like to fallback to  "build googletest if not found in the system"

CMakeLists.txt
Mizux: EDIT:
1. DONE  (we now target cmake 3.13)
2. googletest v1.11 (June 2021) should provide alias -> Can we suppose super build use at least googletest 1.11 ?

PR #312, opened 2023-06-14
Add Riscv32 and Riscv64 Bazel build rules.

prashanthswami:
The CMake builds are already supported, this just forwards those rules into the Bazel BUILD.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,7 +1,7 @@
 # cpu_features, a cross platform C99 library to get cpu features at runtime.
 
 load("@bazel_skylib//lib:selects.bzl", "selects")
-load("//:bazel/platforms.bzl", "PLATFORM_CPU_ARM", "PLATFORM_CPU_ARM64", "PLATFORM_CPU_MIPS", "PLATFORM_CPU_PPC", "PLATFORM_CPU_X86_64")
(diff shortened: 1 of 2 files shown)


Mizux:
FYI: `riscv64` has been added in @platform 0.0.4
https://github.com/bazelbuild/platforms/commit/44d70196ba71780ea4cbebb2f62b53a8862dec76

There is also a `riscv32`: (in @platform 0.0.5)
https://github.com/bazelbuild/platforms/commit/9fd823ec94994e9bd1f4d8fe31ab1f6bc9f81314

@chatelet should we explicitly pull one version in workspace ?
e.g. https://github.com/google/or-tools/blob/24ea806809ca62e98401954a3c43b406eb337c4e/WORKSPACE#L29-L34
note: I don't know wihch version is shipped with each bazel version...

prashanthswami:
Should I add support for RISCV32 as well? I was a bit hyper-focused on RISCV64, but I'm happy to update my CL to account for this.

Mizux:
To support both risc platform I think we may need some code layout rework since currently both are in the same file but we may need two separate files (like we did for arm and aarch64). Need to think about it.

note: Can also be done in a subsequent PR when introducing riscv32 i.e. keep this PR focus on riscv64 bazel support

gchatelet:
Let's do internal cleanup in subsequent patches. Thx @Mizux for having a look.
We may also need to update build/platform as you mentioned. I'll check this later today.

PR #317, opened 2023-08-27
Fix windows x86 and arm64 build and add an option to enable/disable target list_cpu_features

xiaozhuai:
See also: https://github.com/microsoft/vcpkg/pull/33404

I thought this issue should be solved upstream, so I opened this PR.


CMakeLists.txt
```diff
@@ -66,11 +66,11 @@ set(PROCESSOR_IS_LOONGARCH FALSE)
 
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "^mips")
   set(PROCESSOR_IS_MIPS TRUE)
-elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64|ARM64)")
   set(PROCESSOR_IS_AARCH64 TRUE)
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^arm")
   set(PROCESSOR_IS_ARM TRUE)
-elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(x86_64)|(AMD64|amd64)|(^i.86$)")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(x86_64)|(AMD64|amd64)|(^i.86$)")
```
Mizux, line 73: are you sure is not `X86` instead ?

ref: /!\ only found .Net ref -> I supposed they use the same case...
https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.processorarchitecture?view=msbuild-17-netcore
https://learn.microsoft.com/en-us/dotnet/api/system.reflection.processorarchitecture?view=net-7.0

xiaozhuai: See https://learn.microsoft.com/zh-cn/windows/win32/winprog64/wow64-implementation-details?redirectedfrom=MSDN
The `CMAKE_SYSTEM_PROCESSOR` is as same as `PROCESSOR_ARCHITECTURE` on windows.

gchatelet:
Thx for the PR @xiaozhuai !

PR #326, opened 2023-08-31
fix generate_badges.d

Mizux:
usage:
```sh
(cd scripts && ./generate_badges.d)
```

note: `./scripts/generate_badges.d` will fail


gchatelet (approved):

gchatelet:
Thx for the fix :)

PR #332, opened 2023-09-02
Add Intel AMX-FP16 detection

toor1245:
https://cdrdv2-public.intel.com/774990/architecture-instruction-set-extensions-programming-reference.pdf
this instruction set will be in future Granite Rapids (~2024-2025)

image

llvm https://github.com/llvm/llvm-project/blob/9f3e3efd98a29eb8df9e3ad43a573c9141d1ace2/clang/lib/Basic/Targets/X86.cpp#L834


gchatelet:
I assume we don't yet have a CPUID dump at that point correct?

PR #343, opened 2023-09-21
Remove unused function

gchatelet:
This is causing the build to fail with stricter warnings.
`error: no previous prototype for function 'CpuFeatures_IsHwCapCpuidSupported' [-Werror,-Wmissing-prototypes]`


Mizux (approved):
LGTM

PR #348, opened 2024-02-18
Add ECV flag on macOS/iOS

mscdex:
This flag comes directly from [Apple's own page on sysctl names](https://developer.apple.com/documentation/kernel/1387446-sysctlbyname/determining_instruction_set_characteristics).


mscdex:
Erm, well, it doesn't seem like I'm able to sign the CLA because Google doesn't allow me to select which email address(es) I'd like to use when setting up the agreement -- it's either all or nothing.

mscdex:
Also, I'm not concerned about attribution or anything so feel free to make the changes manually, it's no big deal. I just want to be able to drop the rest of the patch I currently have on top of this library.

PR #353, opened 2024-06-18
fixed bug for windows-arm64 (introduced in pull request #291)

NicSavichev:
https://github.com/google/cpu_features/pull/291 introduced shorter INTROSPECTION_TABLE in impl_aarch64_windows.c so while public headers declare AARCH64_LAST_=78 only 54 entries were inited in kGetters kFeatureNames.
Thus iterating all features upto AARCH64_LAST_ caused crash as long as garbage pointer at kGetters[54] was called.
Current implementation seems very unsafe and in order to prevent crash we should check for kGetters array size (that corresponds to INTROSPECTION_TABLE entries). Another solution is static assert for AARCH64_LAST_  == sizeof(kGetters).


toor1245:
@NicSavichev, When this patch was added, there was no common INTROSPECTION_TABLE yet and table was sync with linux implementation, could you try to use `src/impl_aarch64__base_implementation.inl`? 
For example, https://github.com/google/cpu_features/blob/main/src/impl_aarch64_macos_or_iphone.c#L20

PR #356, opened 2024-08-25
Add CI MacOS Arm64 support

toor1245:
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners

Proof of success:
macos-arm64 cmake: https://github.com/toor1245/cpu_features/actions/runs/10551510281

We don't have mock compilation definitions for bazel, so our tests will get real CPU information and the tests will fail:
https://github.com/toor1245/cpu_features/actions/runs/10551510292/job/29229039421#step:6:23

### DIFF
--- a/.github/workflows/aarch64_macos_bazel.yml
+++ b/.github/workflows/aarch64_macos_bazel.yml
@@ -1,0 +1,35 @@
+name: AArch64 MacOS Bazel
+
+on:
+  push:
+  pull_request:
+  schedule:
+    # min hours day(month) month day(week)
+    - cron: '0 0 7,22 * *'
+
+jobs:
+  # Building using the github runner environement directly.
+  bazel:
+    runs-on: macos-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Install Bazel
+        run: |
+          brew update
+          brew unlink bazelisk
+          brew install bazel
+      - name: Check Bazel
+        run: bazel version
+      - name: Build
+        run: >
+          bazel build
+          -c opt
+          --subcommands=true
+          ...
+      - name: Test
+        run: >
+          bazel test
+          -c opt
+          --test_output=errors
+          ...
--- a/.github/workflows/aarch64_macos_cmake.yml
+++ b/.github/workflows/aarch64_macos_cmake.yml
@@ -1,0 +1,43 @@
+name: AArch64 MacOS CMake
+
+on:
+  push:
+  pull_request:
+  schedule:
+    # min hours day(month) month day(week)
+    - cron: '0 0 7,22 * *'
+
+jobs:
+  # Building using the github runner environement directly.
+  xcode:
+    runs-on: macos-latest
+    env:
+      CTEST_OUTPUT_ON_FAILURE: 1
+    steps:
+    - uses: actions/checkout@v2
+    - name: Check cmake
+      run: cmake --version
+    - name: Configure
+      run: cmake -S. -Bbuild -G "Xcode" -DCMAKE_CONFIGURATION_TYPES=Release
+    - name: Build
+      run: cmake --build build --config Release --target ALL_BUILD -v
+    - name: Test
+      run: cmake --build build --config Release --target RUN_TESTS -v
+    - name: Install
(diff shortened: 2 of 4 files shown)


toor1245:
@gchatelet, @Mizux, this PR is ready to review please take a look.

.github/workflows/aarch64_macos_bazel.yml
```diff
@@ -0,0 +1,35 @@
+name: AArch64 MacOS Bazel
```
Mizux, line 1: Can you add this line on top ?
```
# ref: https://github.com/actions/runner-images
```

e.g. https://github.com/google/or-tools/blob/main/.github/workflows/amd64_linux_cmake_cpp.yml#L1

.github/workflows/aarch64_macos_cmake.yml
```diff
@@ -0,0 +1,43 @@
+name: AArch64 MacOS CMake
```
Mizux, line 1: same here
```
# ref: https://github.com/actions/runner-images
```

Mizux:
Requested change: `aarch64` -> `arm64`

Context
for **ARCH** naming I usually use what **Docker** use:
see: https://github.com/docker-library/official-images
Which is based on the ["indexes" in the **OCI** image specification](https://github.com/opencontainers/image-spec/blob/v1.0.0/image-index.md)
Which is based on **Golang GOARCH** naming: https://go.dev/doc/install/source#environment
i.e. `arm64`

For **OS**  we prefer to use macos (as now) following the github runner convention
ref: https://github.com/actions/runner-images
note: github also tend to use `arm` for M1 based runners...

So could you rename file as `arm64_macos_[bazel/cmake].yml` and use `Arm64 MacOS ...` in `name:` (first line) ?

.github/workflows/aarch64_macos_bazel.yml
toor1245: done

.github/workflows/arm64_macos_bazel.yml
```diff
@@ -0,0 +1,36 @@
...
+  push:
+  pull_request:
+  schedule:
+    # min hours day(month) month day(week)
+    - cron: '0 0 7,22 * *'
+
+jobs:
+  # Building using the github runner environement directly.
+  bazel:
+    runs-on: macos-latest
```
gchatelet, line 14: Can you add `# Using M1 processors, ref: https://github.com/actions/runner-images`

.github/workflows/arm64_macos_cmake.yml
```diff
@@ -0,0 +1,44 @@
...
+  push:
+  pull_request:
+  schedule:
+    # min hours day(month) month day(week)
+    - cron: '0 0 7,22 * *'
+
+jobs:
+  # Building using the github runner environement directly.
+  xcode:
+    runs-on: macos-latest
```
gchatelet, line 14: ditto

gchatelet (changes requested):

Mizux:
> Updated, should we rename `aarch64_linux_[bazel/cmake].yml` to `arm64_linux_[bazel/cmake].yml` for consistency?

TLDR: Not yet.

Context:
Currently our top README.md table mix native runner build (windows-latest, macos-13, ubuntu-latest...) and cross compilation using a CMake toolchain (aarch64, mips, powerpc) (+ QEMU for testing when possible).

I think we should split our CI matrix result in two tables:
* native runner build and
* cross-compilation build 
note: currently most of our cells doesn't make any sense e.g Riscv-Windows, mips-macos etc... 

Also to ease the work of @gchatelet to adapt his script to generate the badges, we are thinking to add a prefix `cross_` or `xxx_toolchain` (idea welcome) to be able to automatically split .yml files in two categories then generate both tables (native build and cross compile ones) automatically with few lines change in the script....
ref: https://github.com/google/cpu_features/blob/main/scripts/generate_badges.d

.github/workflows/arm64_macos_bazel.yml
toor1245: done

gchatelet (approved):

PR #359, opened 2024-08-26
Update AArch64 features to Linux 6.10.6

toor1245:
https://elixir.bootlin.com/linux/v6.10.6/source/arch/arm64/include/asm/hwcap.h#L142
https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/hwcap.h#L107-L124
https://developer.arm.com/documentation/ddi0601/2024-06/AArch64-Registers


gchatelet:
Thx @toor1245 !

PR #360, opened 2024-08-27
Remove CI support for FreeBSD

gchatelet:
We used to use Vagrant for FreeBSD continuous integration platforms but the service slowly became unusable.


Mizux:
While the CI job is failing (likely a runner ecosystem issue), I'm not sure the CI support (using Vagrant + VirtualBox) is broken. So before removing everything I would just disable the job and verify if the script is working locally.

Then in a later time we could use an ubuntu-latest base image runner to reenable the job instead...
* https://github.com/actions/runner-images/issues/8730
  * https://github.com/metomi/metomi-vms/pull/34/files

Mizux (approved):
LGTM

PR #361, opened 2024-08-27
Fix wrong arm neon detection

gchatelet:
ARM Architecture Reference Manual suggests to check for the presence of the Advanced SIMD instructions via `__ARM_NEON`


Mizux (approved):
LGTM

for the record:
> `<arm_neon.h>` is provided to define the NEON intrinsics. As these intrinsics are in the user namespace, an
implementation would not normally define them until the header is included. The `__ARM_NEON` macro should be
tested before including the header:
```c
 #ifdef __ARM_NEON
 #include <arm_neon.h>
 #endif /* __ARM_NEON */
 ```

ref: https://developer.arm.com/documentation/ihi0053/d/?lang=en

PR #362, opened 2024-09-01
FreeBSD CI support

toor1245:
Checked another solutions to support FreeBSD CI:
* freebsd-vm: https://github.com/vmactions/freebsd-vm
* cross-platform-actions: https://github.com/cross-platform-actions/action

### freebsd-vm:
It looks like this solution supports only amd64 architecture
https://github.com/vmactions/freebsd-builder/blob/main/conf/freebsd-14.1.conf#L3

However, in builder script there is `VM_ARCH`
https://github.com/vmactions/freebsd-builder/blob/f799d8fb9b84977ec8fa020178262de39eaa4b83/build.sh#L35

and [upcoming features](https://github.com/vmactions/freebsd-vm?tab=readme-ov-file#upcoming-features):
* Runs on MacOS to use cpu accelaration.
* Support ARM and other architecture.

Maybe it’s possible manually configure the architecture somehow, but at the moment it looks like a hardcoded architecture amd64

### cross-platform-actions:

This action supports FreeBSD `amd64` and `arm64` architectures and we support these architectures for FreeBSD:
https://github.com/cross-platform-actions/action?tab=readme-ov-file#freebsd-freebsd

I used this solution, proof of success:
* FreeBSD amd64: https://github.com/toor1245/cpu_features/actions/runs/10652248435/job/29525815775#step:3:556
* FreeBSD arm64: https://github.com/toor1245/cpu_features/actions/runs/10652474246/job/29526357627#step:3:568

### DIFF
--- a/.github/workflows/amd64_freebsd_cmake.yml
+++ b/.github/workflows/amd64_freebsd_cmake.yml
@@ -1,0 +1,31 @@
+name: amd64 FreeBSD CMake
+
+on:
+  push:
+  pull_request:
+  schedule:
+    # min hours day(month) month day(week)
+    - cron: '0 0 7,22 * *'
+
+jobs:
+  make:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build and run tests
+        uses: cross-platform-actions/action@v0.25.0
+        with:
+          operating_system: freebsd
+          architecture: x86-64
+          version: '14.1'
+          shell: bash
+          memory: 5G
+          cpu_count: 4
+          run: |
+            sudo pkg install -y cmake git
+            sudo cmake --version
+            sudo cmake -S. -Bbuild -DCMAKE_BUILD_TYPE=Release
+            sudo cmake --build build --target all -v
+            sudo cmake --build build --target test -v
+            sudo cmake --build build --target install -v
+           
--- a/.github/workflows/arm64_freebsd_cmake.yml
+++ b/.github/workflows/arm64_freebsd_cmake.yml
@@ -1,0 +1,31 @@
+name: Arm64 FreeBSD CMake
+
+on:
+  push:
+  pull_request:
+  schedule:
+    # min hours day(month) month day(week)
+    - cron: '0 0 7,22 * *'
+
+jobs:
+  make:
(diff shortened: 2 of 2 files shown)


toor1245:
@gchatelet, @Mizux, could you please review PR?

Mizux:
Sorry for the delay, still need to think about it...
Please see below my few pro and cons, i.e. my current thought about this PR:

On the bright side:
a) having two jobs working is better than nothing (or our previous broken one) -> Many thanks for taking the time to fix our BSD CI support !

b) leveraging github actions reduce our technical debt and reducing scripts length...

On the cons side: 
c) Using actions means the migration/fork to another CVS (e.g. gitlab) make this more difficult, usually I prefer so stick with the bare minimum actions (i.e. checkout) and only from github (i.e. kind of "official" actions) then having everything in the .yml directly or inside a versioned script. i.e. minimize the vendor lock-in.

d) Using actions means we can't easily locally test or we need to maintain a second scripts with overlap to the github action...
-> I prefer to have a script to run locally and the CI is just a dumb runner which can run it for me.
-> I think, github actions often hide the two bash script lines needed to run it locally.
-> I'm biased with my Makefile + Docker/Vagrant CI setup ;)

e) vmactions/freebsd-builder seems brittle and may allow some supply chain attack (ed very low risk since we don't deploy or share/use any token). Actually, my main concern is more the **test of time**, AFAIK only "neil" seems to be the maintainer of this tooling and I don't like this kind of change where both repos need to be in sync...:
<https://github.com/vmactions/freebsd-builder/commit/3db1a35a5f84167fb47912440ba4e4c893798d43>
<https://github.com/vmactions/vbox/commits/main/> (only 2 stars -> not battle tested and brittle ?)

ps: Still want to explore `vagrant` on `ubuntu-latest` path too (in this case we could keep our ci/Makefile/Vagrant stuff just change the runner base image and do few setup)!
see: https://github.com/magma/magma/pull/15381/files#diff-6cea9de55c2b40b7e046ce2f3b096f80f3f7018ac52fa6e55ba11a6ffce4c6f7R105-R108

As usual feel free to comment any points and change my mind, I'm still thinking about it !
One scenario could be to accept this PR as "temporary" solution until we have a new PR challenging it ;)

PR #366, opened 2024-09-13
Move `c99` flag from BUILD file to bazel cmd line

gchatelet:
Requiring c99 for everyone breaks downstream users.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,9 +3,6 @@
 # Use sparingly for things common to all compilers and platforms.
 ###############################################################################
 
-# Enable Bzlmod for every Bazel command
-common --enable_bzlmod
-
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@ out/
 
 # Bazel artifacts
 **/bazel-*
+MODULE.bazel.lock
```
Mizux, line 17: should be part of the PR365 not here...

gchatelet: Whether you use bzlmod or not, new versions of bazel will create this file.
I agree that it doesn't have to be part of this patch though so I'll commit it at head and sync.

Mizux (approved):

PR #368, opened 2024-10-05
fix RISC-V _Z extension parsing, they don't have a defined order

camel-cdr:
I noticed that the current RISC-V ISA string parsing code assumes there is a order for the extensions. While this is true for single letter extensions, the others, e.g. with _Z prefix, don't have a defined order.


camel-cdr:
How do I sign the CLA without using an google account?

PR #371, opened 2024-11-08
Bump actions/checkout to v4

tinyboxvk:
Resolves node.js deprecation warnings in Actions

`The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2`
`The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v2`
`The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v3`


Mizux (approved):

gchatelet:
thx @tinyboxvk !

PR #372, opened 2024-12-07
cmake: Remove useless check for CMP0077

gruenich:
cpu_features requires CMake 3.13 which sets CMP0077 always to NEW.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,5 @@
 cmake_minimum_required(VERSION 3.13)
 
-# option() honors normal variables.
-# see: https://cmake.org/cmake/help/git-stage/policy/CMP0077.html
-if(POLICY CMP0077)
-  cmake_policy(SET CMP0077 NEW)
-endif()
-
 project(CpuFeatures VERSION 0.9.0 LANGUAGES C)
 
 set(CMAKE_C_STANDARD 99)


Mizux (approved):
LGTM

Xottab-DUTY:
> CMake 3.13 which sets CMP0077 always to NEW.

@gruenich, @Mizux, @gchatelet, this is actually not true.

CMake documentation [states](https://cmake.org/cmake/help/latest/policy/CMP0077.html):
> This policy was introduced in CMake version 3.13. It may be set by [cmake_policy()](https://cmake.org/cmake/help/latest/command/cmake_policy.html#command:cmake_policy) or [cmake_minimum_required()](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required). If it is not set, **CMake warns, and uses OLD behavior**.

So, if we want the NEW behaviour, we need to set the policy to NEW, because the default is OLD.

Mizux:
since our CMake require 3.13 the "new" behaviour is the default, the warning and old behaviour is if you target an older cmake version BUT use a new "cmake" binary... 

https://github.com/google/cpu_features/blob/7d06d4106466d9c600dfc979d7ecf2fa27488d50/CMakeLists.txt#L1

Xottab-DUTY:
Ahhh, I see now. Thanks for the explanation and sorry for bothering!

PR #376, opened 2024-12-11
Add Bazel windows support

Mizux:


gchatelet (approved):
We'd need to update the README.md I think but that can be done afterward.

Mizux:
> We'd need to update the README.md I think but that can be done afterward.

DONE !

PR #377, opened 2024-12-11
Fix ci

Mizux:
Bump bootlin toolchains and fix all jobs
Also Bump qemu to 9.0.4
note: Lastest QEMU release (9.1.2) will break aarch[64]be jobs
see: https://gitlab.com/qemu-project/qemu/-/issues/2580

### DIFF
--- a/scripts/run_integration.sh
+++ b/scripts/run_integration.sh
@@ -33,7 +33,7 @@ function install_qemu() {
     >&2 echo 'QEMU is disabled !'
     return 0
   fi
-  local -r QEMU_VERSION=${QEMU_VERSION:=9.0.2}
+  local -r QEMU_VERSION=${QEMU_VERSION:=9.0.4}
   local -r QEMU_TARGET=${QEMU_ARCH}-linux-user
 
   if echo "${QEMU_VERSION} ${QEMU_TARGET}" | cmp --silent "${QEMU_INSTALL}/.build" -; then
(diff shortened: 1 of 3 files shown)


scripts/run_integration.sh
```diff
@@ -141,38 +141,62 @@ function expand_bootlin_config() {
   local -r GCC_DIR=${ARCHIVE_DIR}/${GCC_RELATIVE_DIR}
 
   case "${TARGET}" in
-    "aarch64")
-      local -r TOOLCHAIN_URL="https://toolchains.bootlin.com/downloads/releases/toolchains/aarch64/tarballs/aarch64--glibc--stable-2021.11-1.tar.bz2"
+    "arm" | "armv7-eabihf")
+      local -r TOOLCHAIN_URL="https://toolchains.bootlin.com/downloads/releases/toolchains/armv7-eabihf/tarballs/armv7-eabihf--glibc--stable-2024.05-1.tar.xz"
```
gchatelet, line 145: [nit] Maybe use a function to create the url?

Mizux: at least a macro for the URL prefix, on my way...

gchatelet: SGTM I don't have a strong opinion on this, copy / paste is OK.

Mizux: see offlline: Better to keep full url since it'd be less error prone to copy/paste them from "copy link address" from 
https://toolchains.bootlin.com

PR #380, opened 2024-12-17
Includes of gtest with < > instead of " ".

gruenich:
Found by Cppcheck (missingInclude).

### DIFF
--- a/test/bit_utils_test.cc
+++ b/test/bit_utils_test.cc
@@ -14,7 +14,7 @@
 
 #include "internal/bit_utils.h"
 
-#include "gtest/gtest.h"
+#include <gtest/gtest.h>
 
 namespace cpu_features {
 namespace {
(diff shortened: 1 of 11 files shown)


gruenich:
Can I spark your interested in this humble improvement, @gchatelet ?

Mizux:
AFAIK bazel use `-I` and **not** `-isystem` so using `<...>` should break the bazel based build

gruenich:
It should not matter to be included by `-I` or `-isystem`, the difference is only that the latter suppresses warnings from those includes.

The main difference is that according to POSIX qutation marks leads to a search relative to the file while the bracets searche relative to the includes (`-I`), see  
https://stackoverflow.com/a/11576616/2799037

gchatelet:
Indeed it should not matter to use `-I` or `-isystem` but if we are to use `<...>` then we also need to move the include above the `"..."` includes.

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

PR #381, opened 2024-12-18
[bazel] inline platforms and remove `platforms.bzl`

gchatelet:
note: replace `PLATFORM_CPU_X86_64` by `"@platforms//cpu:x86_32", "@platforms//cpu:x86_64"`


Mizux (approved):
LGTM

PR #383, opened 2025-02-08
Run GitHub Workflow aarch64_linux_bazel on an ARM processor

cclauss:
[Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) --> `ubuntu-24.04-arm`

Accelerate the longest running GitHub Action job.

Currently, the `aarch64_linux_bazel` job takes three times longer to run than the other GitHub Actions jobs.

### DIFF
--- a/.github/workflows/aarch64_linux_bazel.yml
+++ b/.github/workflows/aarch64_linux_bazel.yml
@@ -8,18 +8,12 @@ on:
     - cron: '0 0 7,22 * *'
 
 jobs:
-  # Building using the github runner environement directly.
+  # Building using the GitHub runner environement directly.
   bazel:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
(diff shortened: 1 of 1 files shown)


gchatelet:
@Mizux WDYT?

Mizux:
this `ubuntu-24.04-arm` runner are still in "Public preview" so we may have to revert this atomic commit but we can still apply it now and revert if needed.

Otherwise the patch seems good and looking at the trace seems to behave as expected (moving from ~10min build to 2min and correctly pull from ubuntu arm64 feed) -> OK to merge IMHO.

Mizux (approved):
LGTM

gchatelet:
@cclauss can I ask you to rebase the patch?

Mizux (approved):

PR #395, opened 2025-05-06
Fix scripts/make_release.sh

gchatelet:
- Make sure to commit `MODULE.bazel`
- Fail before tagging if the repo has unstaged or uncommited changes.


Mizux (approved):
LGTM

Issue #393, opened 2025-05-04
error: #error C++ versions less than C++17 are not supported.


vt-alt:
When building latest release v0.10.0, there is build error:
```
23:02:14 [ 73%] Built target list_cpu_features
23:02:14 gmake[1]: Leaving directory '/usr/src/RPM/BUILD/cpu_features-0.10.0/x86_64-alt-linux'
23:02:14 gmake[2]: Entering directory '/usr/src/RPM/BUILD/cpu_features-0.10.0/x86_64-alt-linux'
23:02:14 [ 55%] Building CXX object test/CMakeFiles/string_view_test.dir/string_view_test.cc.o
23:02:14 cd /usr/src/RPM/BUILD/cpu_features-0.10.0/x86_64-alt-linux/test && /usr/bin/c++ -DCPU_FEATURES_TEST -I/usr/src/RPM/BUILD/cpu_features-0.10.0/test/../include -pipe -frecord-gcc-switches -Wall -fdiagnostics-color=always -g -O2 -flto=auto -O2 -DNDEBUG -std=c++14 -fPIE -MD -MT test/CMakeFiles/string_view_test.dir/string_view_test.cc.o -MF CMakeFiles/string_view_test.dir/string_view_test.cc.o.d -o CMakeFiles/string_view_test.dir/string_view_test.cc.o -c /usr/src/RPM/BUILD/cpu_features-0.10.0/test/string_view_test.cc
23:02:14 In file included from /usr/include/gtest/gtest-message.h:57,
23:02:14                  from /usr/include/gtest/gtest-assertion-result.h:46,
23:02:14                  from /usr/include/gtest/gtest.h:63,
23:02:14                  from /usr/src/RPM/BUILD/cpu_features-0.10.0/test/string_view_test.cc:17:
23:02:14 /usr/include/gtest/internal/gtest-port.h:273:2: error: #error C++ versions less than C++17 are not supported.
23:02:14   273 | #error C++ versions less than C++17 are not supported.
23:02:14       |  ^~~~~
23:02:14 In file included from /usr/include/gtest/gtest.h:67:
23:02:14 /usr/include/gtest/gtest-param-test.h:483:56: error: missing template arguments before '(' token
23:02:14   483 |           typename StdFunction = decltype(std::function(std::declval<Func>()))>
23:02:14       |                                                        ^
23:02:14 /usr/include/gtest/gtest-param-test.h:493:56: error: missing template arguments before '(' token
[...]

PR #396, opened 2025-05-06
Change test c++ standard handling

gchatelet:
Moving from "mandatory c++14" to "at least c++14".


gchatelet:
@vt-alt would you mind trying this patch?
Issue #393 .

PR #397, opened 2025-05-17
Add Arm64 Windows CI support

toor1245:
Github actions added support Windows Arm64 for public repositories with runner name `windows-11-arm`:

- [Windows arm64 hosted runners now available in public preview](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/)
- [Windows 11 by Arm Limited installed packages](https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md)

Cmake has known issue related to build for Windows Arm64. `CMAKE_SYSTEM_PROCESSOR` on Arm64 target shows `AMD64`...

- https://github.com/toor1245/cpu_features/actions/runs/15084083163/job/42404469826#step:3:14

Issue discussion:
- https://gitlab.kitware.com/cmake/cmake/-/issues/15170
- https://gitlab.kitware.com/cmake/cmake/-/issues/25466

So `CMAKE_<LANG>_COMPILER_ARCHITECTURE_ID` must be used to set  `PROCESSOR_IS_AARCH64` architecture:
- https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html
- https://cmake.org/cmake/help/v3.16/variable/CMAKE_LANG_COMPILER_ARCHITECTURE_ID.html

Also, currently we have build fail for windows Arm64, I think it should be fixed as separate patch:
- One of the solution is to add check - https://github.com/google/cpu_features/pull/353
- Another way is just copy `LINE` features from [src/impl_aarch64__base_implementation.inl#L31](https://github.com/google/cpu_features/blob/main/src/impl_aarch64__base_implementation.inl#L31)
- just to remind the root cause maybe we will find universal solution, https://github.com/google/cpu_features/pull/291 :

We can't use "define_introspection_and_hwcaps.inl" as a common file for all operating systems due to msvc compiler error C2099: initializer is not a constant, so as a workaround for Windows I used separate "define_introspection.inl"


toor1245:
@gchatelet, @Mizux, PR is ready to review

.github/workflows/arm64_windows_cmake.yml
```diff
@@ -0,0 +1,26 @@
+# ref: https://github.com/actions/runner-images
```
Mizux, line 1: https://github.com/actions/partner-runner-images

.github/workflows/arm64_windows_cmake.yml
```diff
@@ -0,0 +1,26 @@
...
+name: Arm64 Windows CMake
+
+on:
+  push:
+  pull_request:
+  schedule:
+    # min hours day(month) month day(week)
+    - cron: '0 0 7,22 * *'
```
Mizux, line 9: can we add `workflow_dispatch:` too to be able to run it manually ?

PR #446, opened 2026-04-18
Add some RISC-V options

michael-roe:
This adds the RISC-V extensions for bit manipulation and half-word floating point.


michael-roe:
Sorry, I had a different email address in the commit at first and had to do an amend/force push to change the email to the one you have on the list of contributors

PR #448, opened 2026-04-19
Zig: upgrade 0.16.0

inge4pres:
### Reason for this PR

Zig 0.16.0 was released last week.
Upgrade the build with Zi to the latest stable version.

### Details

* Fix modules methods
* Pin Zig version when installing via mlugg/setup-zig action (since using `latest` can move to the next release without notice)


fiban123:
please merge this, zig 0.16.0 is completely broken with this library right now

PR #455, opened 2026-06-11
ci: add least-privilege read-only token permissions (22 workflows)

Alb3e3:
## Summary

CI workflows are missing top-level \`permissions\` declarations and reference actions by mutable tags/branch names — both are supply-chain risk vectors.

**Changes:**
- Add \`permissions: read-all\` at the workflow level (least-privilege default for all jobs)
- Pin all third-party action references to full commit SHAs

Mutable action references (\`@v4\`, \`@master\`, etc.) allow a compromised upstream to inject arbitrary code into your CI. Pinning to a commit SHA ensures only the audited revision runs regardless of upstream changes.

### DIFF
--- a/.github/workflows/aarch64_linux_bazel.yml
+++ b/.github/workflows/aarch64_linux_bazel.yml
@@ -7,6 +7,9 @@ on:
     # min hours day(month) month day(week)
     - cron: '0 0 7,22 * *'
 
+permissions:
+  contents: read
+
 jobs:
   # Building using the GitHub runner environement directly.
   bazel:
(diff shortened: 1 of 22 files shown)


Alb3e3:
I've rebased this onto `main` and narrowed the change to **permissions-only**.

The previous revision also pinned action versions to commit SHAs, which conflicted with this repo's Dependabot-managed version tags (the recent `actions/checkout` v6→v7 bump was the direct cause of the merge conflict). Since re-pinning would just re-introduce that churn on every Dependabot update, I dropped the SHA pins and kept only the security improvement that doesn't fight your existing workflow: a top-level `permissions: { contents: read }` block on the 22 workflows that lacked one, so the default `GITHUB_TOKEN` is read-only.

No behaviour changes; the conflict is resolved and `zizmor` still passes.

gchatelet:
Thx for the PR @Alb3e3

Issue #457, opened 2026-06-19
Undefined Behavior (Alignment Error) in `SetVendor`


mhansen:
Hi, I wrote some of this using AI, but I stand behind it.

There is an undefined behavior (alignment violation) in `src/impl_x86__base_implementation.inl` when writing the CPU vendor string in the `SetVendor` function.

In `SetVendor`, a `char*` pointer (which is not guaranteed to be 4-byte aligned) is cast to `uint32_t*` and dereferenced for writing.

According to the **C99 Standard (Section 6.3.2.3, paragraph 7)**:
> "A pointer to an object or incomplete type may be converted to a pointer to a different object or incomplete type. If the resulting pointer is not correctly aligned for the pointed-to type, the behavior is undefined."

Since `char` has an alignment requirement of 1 byte, and `uint32_t` typically requires 4-byte alignment, casting an arbitrary `char*` to `uint32_t*` and dereferencing it is undefined behavior if the address is not a multiple of 4.

https://github.com/google/cpu_features/blob/81d13c49649f0714dd41fb56bb246398b6584085/src/impl_x86__base_implementation.inl#L240-L245

This code is specific to `x86`/`x86_64` architectures. While x86/x86_64 CPUs generally support unaligned memory access in hardware, it remains **undefined behavior at the language level**. Compilers can optimize code assuming no UB exists, which can lead to unexpected behavior or compiler-generated crashes.

Furthermore, when compiled with UndefinedBehaviorSanitizer (UBSan) enabled, this triggers a runtime error.

This was originally observed as a crash (`SIGILL`) in `GetX86Info` on Android `x86` (32-bit) targets during NDK builds where I had enabled UBSan trap mode.

You can reproduce this behavior in google3 using Blaze. The repro files are located in `experimental/users/markhansen/cpu_features_ub/`.

**`experimental/users/markhansen/cpu_features_ub/repro_set_vendor.c`**:
```c
#include <stdint.h>
#include <string.h>

#include "third_party/cpu_features/src/impl_x86__base_implementation.inl"

// Stubs for OS-specific functions declared in impl_x86__base_implementation.inl
// but defined in OS-specific .c files.
static void OverrideOsPreserves(OsPreserves* os_preserves) {
  (void)os_preserves;
}
static void DetectFeaturesFromOs(X86Info* info, X86Features* features) {
  (void)info;
  (void)features;
}

int main() {
  Leaf leaf = {0, 0x756E6547, 0x6C65746E, 0x49656E69};
[...]

PR #458, opened 2026-06-19
Avoid undefined behaviour conversions from unaligned char* to int*

### LINKED ISSUE #456: Undefined Behavior (Alignment Error) in `IsVendor`
Hi, I wrote some of this using AI, but I've checked it and stand behind it.

There is an undefined behavior (alignment violation) in `src/impl_x86__base_implementation.inl` when parsing and comparing the CPU vendor string in the `IsVendor` function.

In `IsVendor`, a `const char*` pointer (which is not guaranteed to be 4-byte aligned) is cast to `uint32_t*` and dereferenced.

According to the **C99 Standard (Section 6.3.2.3, paragraph 7)**:
> "A pointer to an object or incomplete type may be converted to a pointer to a different object or incomplete type. If the resulting pointer is not correctly aligned for the pointed-to type, the behavior is undefined."

Since `char` has an alignment requirement of 1 byte, and `uint32_t` typically requires 4-byte alignment, casting an arbitrary `char*` to `uint32_t*` and dereferencing it is undefined behavior if the address is not a multiple of 4.

https://github.com/google/cpu_features/blob/81d13c49649f0714dd41fb56bb246398b6584085/src/impl_x86__base_implementation.inl#L247-L252

This code is specific to `x86`/`x86_64` architectures. While x86/x86_64 CPUs generally support unaligned memory access in hardware, it remains **undefined behavior at the language level**. Compilers can optimize code assuming no UB exists, which can lead to unexpected behavior or compiler-generated crashes.

Furthermore, when compiled with UndefinedBehaviorSanitizer (UBSan) enabled, this triggers a runtime error.

You can reproduce this behavior in google3 using Blaze. The repro files are located in `experimental/users/markhansen/cpu_features_ub/`.

**`experimental/users/markhansen/cpu_features_ub/repro.c`**:
```c
#include <stdint.h>
#include <string.h>

#include "third_party/cpu_features/src/impl_x86__base_implementation.inl"

// Stubs for OS-specific functions declared in impl_x86__base_implementation.inl
// but defined in OS-specific .c files.
static void OverrideOsPreserves(OsPreserves* os_preserves) {
  (void)os_preserves;
}
static void DetectFeaturesFromOs(X86Info* info, X86Features* features) {
  (void)info;
  (void)features;
}

int main() {
  Leaf matching_leaf = {0, 0x756E6547, 0x6C65746E, 0x49656E69};
  char storage[20] __attribute__((aligned(16))) = {0};
  strcpy(storage + 1, "GenuineIntel");
  return IsVendor(matching_leaf, storage + 1);
}
```

Run the repro with Blaze and UBSan diagnostics enabled:
```bash
blaze run --config=ubsan-diagnostics //experimental/users/markhansen/cpu_features_ub:repro
```

Output:
```
src/impl_x86__base_implementation.inl:248:24: runtime error: load of misaligned address 0x7ffffbaac761 for type 'const uint32_t' (aka 'const unsigned int'), which requires 4 byte alignment
0x7ffffbaac761: note: pointer points here
 7f 00 00  aa 47 65 6e 75 69 6e 65  49 6e 74 65 6c 00 aa aa  aa aa aa aa d7 55 00 00  38 7d d8 a4 00
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/impl_x86__base_implementation.inl:248:24
```

Replace the direct pointer casts with `memcpy` to safely handle potentially unaligned memory addresses. Modern compilers (like Clang and GCC) optimize `memcpy` to direct instructions on architectures that support unaligned access (like x86), so there is no performance penalty.

```diff
diff --git a/src/impl_x86__base_implementation.inl b/src/impl_x86__base_implementation.inl
index 0101573..010156f 100644
--- a/src/impl_x86__base_implementation.inl
+++ b/src/impl_x86__base_implementation.inl
@@ -247,9 +247,9 @@
 static int IsVendor(const Leaf leaf, const char* const name) {
-  const uint32_t ebx = *(const uint32_t*)(name);
-  const uint32_t edx = *(const uint32_t*)(name + 4);
-  const uint32_t ecx = *(const uint32_t*)(name + 8);
+  uint32_t ebx, edx, ecx;
+  memcpy(&ebx, name, 4);
+  memcpy(&edx, name + 4, 4);
+  memcpy(&ecx, name + 8, 4);
   return leaf.ebx == ebx && leaf.ecx == ecx && leaf.edx == edx;
 }
```

mhansen:
Fixes #456, #457

### DIFF
--- a/src/impl_x86__base_implementation.inl
+++ b/src/impl_x86__base_implementation.inl
@@ -238,16 +238,17 @@ static bool HasTmmOsXSave(uint32_t xcr0_eax) {
 ////////////////////////////////////////////////////////////////////////////////
 
 static void SetVendor(const Leaf leaf, char* const vendor) {
-  *(uint32_t*)(vendor) = leaf.ebx;
-  *(uint32_t*)(vendor + 4) = leaf.edx;
-  *(uint32_t*)(vendor + 8) = leaf.ecx;
+  memcpy(vendor, &leaf.ebx, 4);
+  memcpy(vendor + 4, &leaf.edx, 4);
+  memcpy(vendor + 8, &leaf.ecx, 4);
   vendor[12] = '\0';
 }
 
 static int IsVendor(const Leaf leaf, const char* const name) {
-  const uint32_t ebx = *(const uint32_t*)(name);
-  const uint32_t edx = *(const uint32_t*)(name + 4);
-  const uint32_t ecx = *(const uint32_t*)(name + 8);
+  uint32_t ebx, edx, ecx;
+  memcpy(&ebx, name, 4);
+  memcpy(&edx, name + 4, 4);
+  memcpy(&ecx, name + 8, 4);
   return leaf.ebx == ebx && leaf.ecx == ecx && leaf.edx == edx;
 }
 


gchatelet (approved):
Thx for the fix. Really appreciate it!

PR #461, opened 2026-06-24
Update bazel module setup

keith:
Downstream in the bazel central registry there is some naive parsing
that doesn't work well with constants like this. We can inline this
value instead. I left around CPU_FEATURES_VERSION to make sure the sed
was unambiguous.

This also removes the `repo_name` which was unused

### DIFF
--- a/scripts/make_release.sh
+++ b/scripts/make_release.sh
@@ -63,7 +63,7 @@ echo -e "${ACTION}Modifying CMakeLists.txt${NOCOLOR}"
 sed -i "s/CpuFeatures VERSION ${LATEST_VERSION}/CpuFeatures VERSION ${VERSION}/g" CMakeLists.txt
 
 echo -e "${ACTION}Modifying MODULE.bazel${NOCOLOR}"
-sed -i "s/CPU_FEATURES_VERSION = \"${LATEST_VERSION}\"/CPU_FEATURES_VERSION = \"${VERSION}\"/g" MODULE.bazel
+sed -i "s/version = \"${LATEST_VERSION}\",  # CPU_FEATURES_VERSION/version = \"${VERSION}\",  # CPU_FEATURES_VERSION/g" MODULE.bazel
 
 echo -e "${ACTION}Modifying build.zig.zon${NOCOLOR}" 
 sed -i "s/.version = \"${LATEST_VERSION}\"/.version = \"${VERSION}\"/g" build.zig.zon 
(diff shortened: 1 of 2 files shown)


keith:
an alternate common pattern we can do is just set this to `version = "0"` and let it be patched in the BCR. this number only matters when being resolved in that setting

Mizux:
Got the same issue with or-tools, highs and few other modules I maintain on BCR, tooling don't support variable in version...

Mizux:
> an alternate common pattern we can do is just set this to `version = "0"` and let it be patched in the BCR. this number only matters when being resolved in that setting

need to provide an overlay or a patch (cumbersome in bcr ecosystem since compared to cmake fetchContent we don't easily have the orig repo cloned) so if we can have the correct version upstream will be better imho

scripts/make_release.sh
```diff
@@ -63,7 +63,7 @@ echo -e "${ACTION}Modifying CMakeLists.txt${NOCOLOR}"
 sed -i "s/CpuFeatures VERSION ${LATEST_VERSION}/CpuFeatures VERSION ${VERSION}/g" CMakeLists.txt
 
 echo -e "${ACTION}Modifying MODULE.bazel${NOCOLOR}"
-sed -i "s/CPU_FEATURES_VERSION = \"${LATEST_VERSION}\"/CPU_FEATURES_VERSION = \"${VERSION}\"/g" MODULE.bazel
+sed -i "s/version = \"${LATEST_VERSION}\",  # CPU_FEATURES_VERSION/version = \"${VERSION}\",  # CPU_FEATURES_VERSION/g" MODULE.bazel
```
aiuto: Perhaps buildozer is the right choice for this?
That seems to be what the auto-publish scripts in other repos use.

gchatelet: I agree that buildozer is probably less brittle, the additional dependency is bearable

Something like
```
buildozer 'set version "${LATEST_VERSION}"' //MODULE.bazel:all
```

keith: updated to that (requires a slightly different target), of course that means buildozer must be installed wherever this runs, is that the case today?

gchatelet: Thx! Did you test it?

I'm fine installing buildozer if it helps. No big deal.

keith: yep it works locally, notably `:all` does not work since it changes every `version =` in the file

gchatelet (approved):

scripts/make_release.sh
keith: it was using the wrong bash variable i think actually, fixed

gchatelet (approved):
Oops good catch

PR #464, opened 2026-07-01
aarch64: fix SME_LUTV2 /proc/cpuinfo flag string (smelutv1 -> smelutv2)

EylonKrause:
## Description

The `AARCH64_SME_LUTV2` entry in the AArch64 feature table uses `"smelutv1"` as its `/proc/cpuinfo` match string, while both its enum name (`smelutv2`) and its HWCAP2 macro (`AARCH64_HWCAP2_SME_LUTV2`) are LUTv2. The Linux kernel prints this feature as `smelutv2` (`KERNEL_HWCAP_SME_LUTV2` in [`arch/arm64/kernel/cpuinfo.c`](https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/cpuinfo.c)), so the `/proc/cpuinfo` string-scan fallback never matches and never sets the `SME_LUTV2` bit. It is the sole string-vs-name mismatch among the SME/FP8 entries.

The `getauxval(AT_HWCAP2)` bit path is unaffected — this only fixes the cpuinfo-scan fallback used when `getauxval` is unavailable.

## Fix

`"smelutv1"` → `"smelutv2"`, matching the kernel token and the entry's own name.

I don't have AArch64 hardware to exercise the cpuinfo path at runtime, but the fix is verifiable against the kernel's hwcap string table (linked above) and the entry's name / HWCAP2 macro.


EylonKrause:

PR #466, opened 2026-07-01
hwcaps: dont return an uninitialized hwcap on elf_aux_info failure

EylonKrause:
## Description

On FreeBSD/OpenBSD, `GetElfHwcapFromElfAuxInfo` reads the ELF hwcaps via `elf_aux_info`:

```c
static unsigned long GetElfHwcapFromElfAuxInfo(int hwcap_type) {
  unsigned long hwcap;
  elf_aux_info(hwcap_type, &hwcap, sizeof(hwcap));
  return hwcap;
}
```

`hwcap` is uninitialized and the `elf_aux_info` return value is ignored. `elf_aux_info` only writes `*buf` on success; when the requested entry is absent it returns non-zero and leaves the buffer **untouched**. `CpuFeatures_GetHardwareCapabilities` requests `AT_HWCAP2` unconditionally, and `AT_HWCAP2` is genuinely absent on some arch/kernel combinations — so on those, this returns an **indeterminate stack value**, which is then bit-tested for CPU features → non-deterministic false detection.

## Fix

Initialize `hwcap = 0` and return 0 when `elf_aux_info` fails, mirroring the Linux `getauxval` path (which returns 0 when the entry is unavailable).

## Testing

FreeBSD/OpenBSD-only path, so not exercised on this Linux host; the fix is a straightforward initialize + return-check (the uninitialized read is clear from the source).

### DIFF
--- a/src/hwcaps_freebsd_or_openbsd.c
+++ b/src/hwcaps_freebsd_or_openbsd.c
@@ -30,7 +30,10 @@ const char* CpuFeatures_GetBasePlatformPointer(void);
 #include <sys/auxv.h>
 
 static unsigned long GetElfHwcapFromElfAuxInfo(int hwcap_type) {
-  unsigned long hwcap;
+  unsigned long hwcap = 0;
+  // elf_aux_info() leaves the output buffer untouched when the requested entry
+  // is absent (e.g. AT_HWCAP2 on some arch/kernel combos), so initializing
+  // hwcap to 0 avoids returning an indeterminate value on failure.
   elf_aux_info(hwcap_type, &hwcap, sizeof(hwcap));
   return hwcap;
 }


EylonKrause:

src/hwcaps_freebsd_or_openbsd.c
```diff
@@ -30,8 +30,11 @@ const char* CpuFeatures_GetBasePlatformPointer(void);
 #include <sys/auxv.h>
 
 static unsigned long GetElfHwcapFromElfAuxInfo(int hwcap_type) {
-  unsigned long hwcap;
-  elf_aux_info(hwcap_type, &hwcap, sizeof(hwcap));
+  unsigned long hwcap = 0;
+  // elf_aux_info() leaves the output buffer untouched and returns non-zero when
+  // the requested entry is absent (e.g. AT_HWCAP2 on some arch/kernel combos),
+  // so return 0 instead of an indeterminate value.
+  if (elf_aux_info(hwcap_type, &hwcap, sizeof(hwcap)) != 0) return 0;
```
Mizux: if elf_aux_info let the buffer untouched then simply returning hwcap (initialised to 0) is sufficient ?

EylonKrause:
Good idea — simplified to rely on the zero-initialized `hwcap`. Since `elf_aux_info()` leaves the output buffer untouched when the entry is absent, initializing to 0 and returning it covers the failure case without the explicit return-value check. Updated.

PR #467, opened 2026-07-02
Declare windows_utils.h as a header of cpuinfo on ARM64

fmeum:
`src/impl_aarch64_windows.c` includes `"internal/windows_utils.h"`, but the header is only listed in the `hdrs` of `:cpuinfo` (and `:cpuinfo_for_testing`) in the x86 branch of the `select`. As a result, Bazel builds of `:cpuinfo` for windows-arm64 fail with:

```
src/impl_aarch64_windows.c:24:10: fatal error: 'internal/windows_utils.h' file not found
```

This adds the header to the ARM64 branch of both targets.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -261,6 +261,7 @@ cc_library(
         PLATFORM_CPU_ARM64: [
             "include/cpuinfo_aarch64.h",
             "include/internal/cpuid_aarch64.h",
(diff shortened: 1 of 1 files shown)


Mizux:
https://github.com/google/cpu_features/blob/044ec8c0f785895d312b6c37845ec5d8811c6406/src/impl_aarch64_windows.c#L24

PR #470, opened 2026-08-27
Move googletest to `dev_dependency = True`

UebelAndre:
`googletest` is not needed by core `cpu_features` libraries and brings in additional dependencies. This change moves it to `dev_dependency = True` to avoid this impact on consumers.


UebelAndre:
@gchatelet friendly ping here 🙏

PR #471, opened 2026-09-10
Fix undefined behavior in StringView trimming and validate positive number parsing

jdymitarai:
### Summary of Changes

1. **Fix undefined behavior in StringView trimming**:
   - In standard C (C99 7.4 / C11 7.4), the argument to `<ctype.h>` functions like `isspace()` must be representable as `unsigned char` or equal to `EOF`.
   - When `char` is signed (default on x86/x64 MSVC and GCC/Clang), non-ASCII characters (`>= 0x80`) produce negative integer values when passed to `isspace()`, causing undefined behavior or debug assertions (e.g. `_ASSERTE((unsigned)(c + 1) <= 256)` on MSVC).
   - Cast characters to `(unsigned char)` before invoking `isspace()`.

2. **Reject empty views in `ParsePositiveNumberWithBase`**:
   - Previously, parsing `"0x"` stripped the `"0x"` prefix leaving an empty remainder view (`size == 0`). The loop in `ParsePositiveNumberWithBase` never iterated and returned `result = 0`. As a result, the malformed literal `"0x"` was incorrectly treated as valid integer `0` instead of returning `-1`.
   - Returning `-1` if `view.size == 0` ensures incomplete hex literals without digits properly fail.

3. **Prevent signed integer overflow in `ParsePositiveNumberWithBase`**:
   - Added bounds check (`result > (INT_MAX - value) / base`) before `result = (result * base) + value` to guard against signed integer overflow on oversized strings.

4. **Unit Tests**:
   - Added test cases in `test/string_view_test.cc` for non-ASCII whitespace trimming, `"0"`, `"0x0"`, `"0x"`, malformed hex suffixes, and overflow inputs.

### Verification
All 17 unit tests in `string_view_test` and all CMake test suites pass cleanly on MSVC.


jdymitarai:
Closing this PR to avoid review overhead on the team. The context remains in the thread for future reference if helpful. Thanks for your time.

PR #473, opened 2026-09-13
fix(x86): require OS AVX support for F16C, VAES and VPCLMULQDQ

dennisimoo:
F16C, VAES and VPCLMULQDQ were reported from their CPUID bits even when the OS
does not enable the required XMM/YMM state. A caller selecting instructions from
these flags could therefore select an unsupported path.

Move the three assignments into the existing `os_preserves->avx_registers`
block alongside AVX and FMA3. This follows the SIMD availability contract at the
top of the implementation; LLVM's [host detection](https://github.com/llvm/llvm-project/blob/main/llvm/lib/TargetParser/Host.cpp)
also gates all three features on `HasAVXSave`. No public API changes.

The regression test advertises all three CPUID features and toggles OS register
support. All three negative assertions fail before the fix; both disabled and
enabled cases pass afterward.

Tested on x86-64 Linux with GCC 13:

```
cmake -S . -B /tmp/cpu-features-build -G Ninja -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build /tmp/cpu-features-build -j 2
ctest --test-dir /tmp/cpu-features-build --output-on-failure
/tmp/cpu-features-build/list_cpu_features --json
```

All four native test executables pass. This was found during source review, not
from a reported downstream incident. Prepared with AI assistance.


dennisimoo:
Fixed formatting in 6fe7b10 using CI’s formatter. All four local tests pass; CI awaits approval.

